Issue #46, opened 2026-07-02
Editor-path performance debt and robustness cleanups


vukefn:
## Problem Statement

The extension does avoidable work on hot editor paths. These issues predate the 0.7 changes but affect every user: the diagnostics listener loads every changed document in the workspace (any language) before checking whether it is a Verse file, and re-reads configuration inside that loop; the CodeLens provider recomputes two whole-file scans once per import line on every refresh (which fires on keystrokes), making refresh cost quadratic in imports; and identifier lookup over the digest index is a linear scan with per-entry string allocation. Two latent robustness smells accompany them: the precompiled-digest availability flag latches off permanently on first failure with no retry path, and the document editor's two import-block filters disagree about empty strings (currently unreachable, but a trap).

## Solution

Cheap, behavior-preserving optimizations on the hot paths: pre-filter by file extension before loading documents, hoist loop-invariant computation and configuration reads, index the digest by lowercased key at load time, make precompiled availability retryable on cache clear, and unify the block filters.

## User Stories

1. As a Verse developer in a mixed-language workspace, I want diagnostics from other languages ignored without loading their documents, so that unrelated projects do not tax the Verse extension.
2. As a Verse developer, I want CodeLens refreshes to scan the file once per refresh, so that typing near an import block stays smooth in large files.
3. As a Verse developer with the experimental digest lookup enabled, I want identifier lookups served from an index, so that unknown-identifier resolution does not degrade with digest size.
4. As a maintainer, I want configuration reads hoisted out of per-URI loops, so that event handlers do minimal work per event.
5. As a maintainer, I want the precompiled digest path to become available again after a transient failure once caches are cleared, so that one bad read does not permanently degrade the session.
6. As a maintainer, I want the two import-block filters to use one predicate, so that a future change to import gating cannot make them diverge observably.

## Implementation Decisions

- The diagnostics listener filters URIs by the Verse file extension (cheap string check) before any document load, and reads configuration once per event, not per URI.
- The CodeLens provider computes the multiple-full-path and multiple-relative flags once before its per-line loop; behavior is identical.
- The digest parser builds a lowercased-key index when digests load; lookups stop allocating per-entry lowercase copies.
- Clearing the digest cache also resets the precompiled-availability flag so the loader re-probes on next use.
- The document editor's block-content filters share one predicate (dropping null and empty), keeping both paths consistent.

## Testing Decisions

[...]

Issue #45, opened 2026-07-02
Unify Verse declaration parsing into one shared parser


vukefn:
## Problem Statement

Five independent regex-based parsers define "what Verse declarations exist" across the codebase: the digest build script, the runtime digest parser, the assets digest parser, the project tree builder, and the import path converter's module-definition detection (twice). They have different coverage (one recognizes interfaces/enums/extension methods, another only public visibility, another only classes/structs) and they have already diverged in a user-visible way: the build script classifies structs correctly while the runtime fallback mistypes them as variables, so import suggestions differ depending on which subsystem answers. Every Verse grammar evolution (UEFN updates add specifiers and forms regularly) must be replicated in five places, and a missed one diverges silently. The build script also re-declares the digest data types instead of importing them, so a one-sided contract change compiles cleanly and breaks at runtime, and three separate hardcoded digest-file lists must be kept in sync by hand.

## Solution

Extract one shared Verse declaration parser that all five consumers use, one shared set of digest data types imported by both producer and consumers, and one digest-file manifest. Consumers keep their own filtering policy (visibility, declaration kinds) but share the extraction.

## User Stories

1. As a maintainer, I want a single module that turns Verse source text into declaration records, so that grammar knowledge lives in one place.
2. As a maintainer, I want each consumer to declare its filtering policy (which kinds, which visibilities) against the shared records, so that intentional differences are explicit configuration rather than accidental regex drift.
3. As a Verse developer, I want structs classified identically whether digest data is precompiled or runtime-parsed, so that suggestions do not depend on internal fallbacks.
4. As a maintainer, I want a new Verse specifier or declaration form supported by editing one parser, so that grammar updates are one change, not five.
5. As a maintainer, I want the digest build script to import the same type definitions the runtime loader deserializes into, so that a contract change on either side fails at compile time.
6. As a maintainer, I want one manifest of digest files consumed by the build script, the precompiled loader, and the runtime fallback, so that adding or renaming a digest is a single edit.
7. As an automated maintenance agent, I want the parser's interface documented and stable, so that future automated changes to consumers do not need to touch parsing.
8. As a maintainer, I want the parser to expose module-stack context (which module a declaration belongs to), so that consumers needing qualified paths do not re-derive nesting.

## Implementation Decisions

- A new shared parsing module (in the services layer) exposes iteration over declaration records: name, kind (module/class/struct/interface/enum/function/variable/extension method), visibility, module path context, and source line.
[...]

Issue #43, opened 2026-07-02
Lifecycle and wiring fixes: shared ImportHandler, assets watcher, timers, regex state


vukefn:
## Problem Statement

Four small, confirmed correctness defects degrade the extension's reliability:

1. Auto-import and quick fix disagree: the diagnostics pipeline constructs its own import handler without the asset-digest dependency and extension context, so "Did you mean A.B.C.ClassName" errors auto-import a path that wrongly includes the class name — the exact bug fixed in 0.6.2 — while the quick-fix menu produces the correct path.
2. The watcher for the external assets digest can never fire (it watches an absolute out-of-workspace path with a plain string pattern, which the editor API documents as unsupported), and the cache lifetime was raised from 30 seconds to 5 minutes — so after adding or renaming assets in UEFN, wrong imports can be produced for up to 5 minutes.
3. Snooze timers leak: invoking the snooze command from the command palette while already snoozed orphans the previous interval timer, and the status bar handler's dispose is never registered, so an active snooze keeps ticking after the extension deactivates — including showing a notification and writing configuration after deactivation. The cache's pending debounce timer is similarly not cleared on disposal.
4. Detection of explicit module definitions reuses one stateful global regex across many files, so whether a module definition is found depends on file iteration order and match offsets — ambiguity detection silently drops valid locations.

## Solution

One hygiene pass: inject the shared, fully-wired import handler everywhere; watch the external digest with the pattern form the API supports for out-of-workspace paths (keeping the longer cache lifetime as a backstop only); make every timer-owning component properly disposable and registered with the extension context, with re-entry guards; and make the module-definition matching stateless.

## User Stories

1. As a Verse developer, I want auto-import and quick fix to produce the same import for the same compiler error, so that the automatic behavior is as trustworthy as the manual one.
2. As a Verse developer, I want asset class names excluded from auto-imported paths, so that the 0.6.2 fix holds on every code path.
3. As a Verse developer, I want newly added or renamed UEFN assets recognized promptly after the digest regenerates, so that suggestions stay correct mid-session.
4. As a Verse developer, I want exactly one snooze countdown running no matter how I start or restart snooze, so that snooze ends once, at the right time.
5. As a Verse developer, I want no notifications or settings writes from the extension after it is disabled or reloaded, so that a deactivated extension is actually inactive.
6. As a Verse developer, I want ambiguous-module detection to find every definition regardless of project file order, so that conversion choices are complete and stable.
[...]

PR #47, opened 2026-07-02
fix: rebuild project path cache lookup with validated location contract

### LINKED ISSUE #41: Rebuild project path cache lookup contract and slim the cache subsystem
## Problem Statement

When the project path cache is enabled (the default) and warm, converting imports between relative and absolute form is broken: "Use Absolute Path" writes malformed `using` statements into the file (dotted cache paths glued into path templates), unrelated identifiers are offered as module locations (substring suffix matching, and class/struct names treated as modules), a spurious ambiguity quick-pick appears for modules that are not ambiguous, and the nearest-module preference that the filesystem scan used to guarantee is lost because a cache hit short-circuits everything including validation.

Separately, the cache subsystem carries structural dead weight that makes it fragile to fix and maintain: the "tree" is never actually a tree (all nodes sit flat under the root while three consumers recurse over it as if nested), three public accessors have no callers, derivable metadata is persisted under a second storage key, the schema version constant is duplicated in two classes, and cache rebuilds parse project files strictly sequentially.

## Solution

Rebuild the cache's lookup surface and internal structure in one pass so that path conversion backed by the cache is indistinguishable from (but faster than) the filesystem scan: same result shape, same locality preference, same precision, validated against the filesystem before use. Internally, replace the fake tree with a flat store plus purpose-built indexes, delete the unused surface, unify the schema version, and parallelize rebuilds.

## User Stories

1. As a Verse developer, I want "Use Absolute Path" to produce a valid absolute import path when the cache is warm, so that my project compiles after conversion.
2. As a Verse developer, I want "Use Relative Path" and "convert all" to apply the same correct path logic, so that bulk conversions are safe.
3. As a Verse developer, I want the ambiguity quick-pick to appear only when a module genuinely exists in multiple locations, so that I am not interrupted for modules with one home.
4. As a Verse developer, I want the module nearest to my current file preferred when several share a name, so that conversions match my project's structure.
5. As a Verse developer, I want class, struct, enum, function, and variable names never offered as module locations, so that conversions cannot target non-modules.
6. As a Verse developer, I want a module named "Utils" to never be matched by "MyUtils", so that lookups are exact rather than substring-based.
7. As a Verse developer, I want stale cache entries verified against the filesystem before being applied, so that a rename or move since the last rebuild cannot produce a wrong import.
8. As a Verse developer, I want the cache rebuild ("Verse: Rebuild Project Path Cache") to finish quickly on large projects, so that I use it instead of avoiding it.
9. As a Verse developer, I want "convert all" on a large file to complete without per-line full-cache walks, so that the editor stays responsive.
10. As a maintainer, I want the cache to expose only methods that have callers, so that the public surface is honest and testable.
11. As a maintainer, I want exactly one schema version constant shared by the builder and the cache validator, so that a version bump cannot half-apply.
12. As a maintainer, I want the persisted format to contain no derivable or always-empty fields, so that storage stays minimal and consistent.
13. As a maintainer, I want lookup logic covered by unit tests, so that future refactors of the cache cannot silently regress path conversion.
14. As an automated maintenance agent, I want the lookup contract documented in the code, so that future automated changes respect it.

## Implementation Decisions

- ProjectPathCache stores a flat collection of declaration records plus two indexes: an identifier-name index (existing) and a new normalized-module-path index for O(1) module lookups.
- The lookup method used by path conversion returns results in the same shape the filesystem scan produces (Content-relative location prefixes), or the converter maps cache records into that shape in one adapter; either way there is exactly one contract, stated in doc comments.
- Lookups filter to module-type records only.
- Matching is segment-boundary exact matching on the normalized module path; no unanchored suffix matching.
- The converter treats cache results as candidates: it validates each against the filesystem (folder existence) and preserves the locality ordering (sibling folder, then ascending directories from the current file) before offering results; if no cache candidate survives validation, the filesystem scan runs as a fallback.
- The unused public accessors (tree getter, identifier lookup, module-contents walk) and the persisted metadata record are deleted; loading gates on the serialized payload alone.
- The cache schema version becomes a single shared constant in the shared cache types module, referenced by both builder and cache.
- The serialized format becomes flat (list of records plus indexes rebuilt on load); a schema version bump invalidates old caches cleanly.
- Cache rebuild parses project files with bounded concurrency instead of strictly sequentially.
- The three cache settings (enable, auto-rebuild, watcher debounce) keep their existing names and semantics.

## Testing Decisions

- Good tests exercise external behavior only: given a set of declaration records and a requested module path, assert which locations are returned, in what order, and in what shape; given cache-plus-filesystem disagreement, assert the filesystem wins.
- Modules under test: the cache lookup (matching, type filtering, boundary precision, ordering) and the converter's location-assembly logic (path template composition for single and multiple locations).
- Serialization round-trip: build, serialize, load, and assert lookups behave identically.
- Prior art: the ImportSuggestionExtractor jest suite (vscode mocked via the shared mock) shows the pattern for pure-logic tests in this repo.

## Out of Scope

- Unifying Verse declaration parsing across the codebase (separate PRD).
- Changes to ImportCodeLensProvider internals (covered by the performance PRD).
- New user-facing settings or commands.
- Changes to how the auto-import pipeline consumes diagnostics.

## Further Notes

Release blocker for v0.7.0: the cache is the release's headline feature and currently corrupts the feature it accelerates. This PRD intentionally includes the structural slim-down (maintainer decision) so the fix lands on the final architecture instead of patching a structure that is then rebuilt post-release.

vukefn:
Closes #41

## What this fixes

With the project path cache enabled (the default), path conversion was broken whenever the cache was warm: `lookupModulePath` returned dot-notation declaration paths that `convertToFullPath` consumed as directory location prefixes (producing paths like `/game@fortnite.com/gameInventory/Inventory`), matching was unanchored suffix over ALL node types (`Utils` matched `MyUtils`; classes offered as modules), and any cache hit short-circuited both the locality-aware folder search and all filesystem validation.

## How

- New pure module `moduleLocationLookup` owns the lookup: exact, segment-boundary, case-sensitive matching over module declarations only. It resolves in-file module nesting (`HUD.Textures` declared in one file) as well as implicit folder segments, and returns locations in the converter's location-prefix contract (`""` / `"/Dir/Sub"`) with the declaring file as evidence.
- `findModuleLocations` now runs the locality-aware Phase 1 folder search first (as before the cache existed), then uses the cache only to accelerate Phase 2 (explicit declarations), validating every candidate's source file against the filesystem and falling back to the full scan when nothing survives.
- Structural slim-down per the PRD: the fake tree is gone (nodes were always flat under root while three consumers recursed). `ProjectPathTreeBuilder` is replaced by `ProjectPathScanner` (flat output, files parsed with bounded concurrency of 8). Dead accessors (`getTree`, `lookupIdentifier`, `getModuleContents`), the derivable persisted metadata, and the duplicated `CACHE_VERSION` constants are deleted; the shared `PROJECT_CACHE_VERSION` lives in the types module and is bumped to `2`, so existing caches rebuild automatically on first use (the legacy metadata storage key is also cleaned up).
- Public surface kept: `getStats()` shape, the three `cache.*` settings, watcher/debounce behavior, and the `ImportPathConverter` constructor are unchanged. `extension.ts` and `CommandsHandler` needed no changes.

## Out of scope (per PRD)

The stateful-regex bug in the Phase 2 scan and timer disposal are #43; the CodeLens provider is #46; declaration-parsing unification is #45.

## Verification

- `npm run compile`: clean
- `npm test`: 3 suites, 24 tests, all passing (17 new lookup tests covering exact/boundary matching, type filtering, nesting, workspace-is-Content normalization, dedup, ordering, and a serialization round-trip; 4 new path-assembly tests; plus the existing extractor suite)


vukefn:
@claude review this PR

PR #154, opened 2026-08-05
ci: trigger CI on merge_group so a merge queue can gate on it

### LINKED ISSUE #152: Enable a merge queue so parallel PRs stop invalidating each other
Three PRs (#147, #148, #149) were open simultaneously today. Merging them showed two separate costs of the current setup.

The Default Branch ruleset sets `strict_required_status_checks_policy: true` ("require branches to be up to date before merging"), so every merge invalidates every other open PR whether or not it actually conflicts. With N open PRs that is O(N^2) in CI reruns and manual branch updates; #147 needed a `Merge remote-tracking branch 'origin/main'` commit purely to satisfy it.

It also does not protect against semantic conflicts: two PRs that touch no common line but still break each other, because each one's CI ran against a main that did not include the other.

A merge queue solves both. It builds each PR against the projected post-merge main and merges only what is green, which is the "not rocket science rule" that Bors, Zuul and Mergify all implement. Free on public repos.

**There is a hard ordering requirement.** Per GitHub's docs, "you must use the `merge_group` event to trigger your GitHub Actions workflow when a pull request is added to a merge queue... the merge will fail as the required status check will not be reported." `ci.yml` currently triggers on `push` and `pull_request` only, so enabling the queue first would produce a queue that hangs forever on a `build` check that never runs.

So this lands in two separately-merged steps:

1. Add `merge_group: types: [checks_requested]` to `ci.yml`, merged on its own, then confirm ordinary PRs still report `build` and `test-integration` (a malformed `on:` block stops a workflow silently rather than erroring).
2. Ruleset: add the `merge_queue` rule with merge method `merge`, promote `test-integration` to a required check, and set `strict_required_status_checks_policy: false` - the queue replaces it and the two fight each other.

Promoting `test-integration` is worth doing here specifically: it runs on every PR today and gates nothing, so the extension-host suite and the packaged-VSIX activation test can both fail into a green merge. A queue is what makes requiring it affordable, because it runs once per queue entry rather than on every rebase.

Acceptance:
- [ ] `ci.yml` reports `build` and `test-integration` on a merge group ref (`gh-readonly-queue/main/...`), not just on PR heads
- [ ] Two PRs queued together both land with neither manually updated
- [ ] Existing rulesets captured to a file before any change

vukefn:
Step 1 of #152. Lands alone, before the ruleset change.

A queued PR is built on a temporary `gh-readonly-queue/main/...` ref, which is neither a `push` nor a `pull_request`. Per GitHub's docs, without a `merge_group` trigger "the merge will fail as the required status check will not be reported" — the queue would wait forever on a `build` that never runs. So the trigger has to be on `main` *before* the `merge_queue` rule exists, not after.

`checks_requested` is the only valid activity type today; naming it explicitly is what the docs recommend so the workflow stays specific if more are added.

No changelog entry: CI plumbing is not user-facing, and `CLAUDE.md` scopes the requirement to changes that are.

## Verification

```
npm run compile      tsc, no errors
npm test             14 suites, 221 tests, all passed
npm run format:check All matched files use Prettier code style!
```

`push` and `pull_request` resolve byte-identically to before; `merge_group` is a correctly-indented sibling. Jobs still resolve to `[build, test-integration]`.

## Review

`verdict: PASS_WITH_SUGGESTIONS`, no Critical findings. Two things worth recording, neither blocking:

**Important, pre-existing, and it falsifies a premise in #152.** `pr-artifact.yml` has a job named `build` — the same string as the sole required check context. PR #149's head SHA carries two check runs called `build`, both from the `github-actions` app, which `integration_id` cannot disambiguate. It does not hang the queue, because `pr-artifact.yml` is `pull_request`-only and a merge group sees exactly one `build`. But it means `build` denotes two things on a PR and one in the queue, and `pr-artifact.yml`'s version runs neither the tests nor the format check. **Worth renaming to `package` before step 2**, not after.

**Suggestion, deliberately not taken.** `merge_group` carries no `branches:` filter while the other two are pinned to `[main]`. `on.merge_group.branches` appears to work but is undocumented (github/docs#32879), and relying on an undocumented filter seemed worse than the asymmetry. Correct as written, since the queue is only being enabled on the default branch.

Closes #152

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,6 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
-  # Required for the merge queue: a queued PR is built on a temporary
-  # gh-readonly-queue/main/... ref, which is neither a push nor a pull_request.
-  # Without this the required checks never report and the queue hangs.
   merge_group:
     types: [checks_requested]
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,11 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  # Required for the merge queue: a queued PR is built on a temporary
+  # gh-readonly-queue/main/... ref, which is neither a push nor a pull_request.
+  # Without this the required checks never report and the queue hangs.
+  merge_group:
+    types: [checks_requested]
 
 jobs:
   build:


.github/workflows/ci.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  # Required for the merge queue: a queued PR is built on a temporary
```
vukefn: remove comment

vukefn: Removed — the diff is the two lines now. The rationale is in the PR body and #152, which is the better place for it anyway.

PR #182, opened 2026-08-06
fix(imports): decide code lens placement with the shared import scanner

### LINKED ISSUE #167: ImportCodeLensProvider scans lines itself, so code lenses still appear on commented-out and module-scoped imports
Found by the review gate on #163 (the fix for #127).

#127 gave `scanModuleImports` block-comment awareness, and every operation in `ImportDocumentEditor` reads through it. `ImportCodeLensProvider` does not — `provideCodeLenses` runs its own per-line loop calling `ImportFormatter.isModuleImport` directly (`src/imports/ImportCodeLensProvider.ts:154`, and again at `:178` and `:211` for the "convert all" detection), so it still treats a commented-out `using` as a live import.

Result: a `using` inside a `<# ... #>` block gets a **Use relative path** / **Use absolute path** lens, and it counts toward the multi-import "convert all" lens. Acting on either lens edits a line inside a comment.

```verse
<#
using { /Old/Path }
#>

code()
```

Two things separate this from #127 and are worth fixing together, since both come from the same cause — the provider not using the shared scanner:

- **Block comments.** The lens offers an action on inert text.
- **Indentation.** `isModuleImport(trimmedLine, nextLine)` is called on the trimmed line with no positional context, so a module-scoped `using` indented inside a `module:` body also gets lenses. `scanModuleImports` rejects those, which is why the editor no longer touches them.

The fix is to route the provider through `scanModuleImports` and map the returned `startLine`/`endLine` spans to lens ranges, rather than re-deriving what counts as an import. That is what the scanner's docstring says it exists for — "the single scanner behind every import-editing operation so they all agree on what counts as an import" — and the provider is the remaining caller that disagrees.

`ImportPathConverter` (`:426`, `:517`) has the same shape and should be checked in the same pass, though its callers may already have narrowed the lines it sees.

vukefn:
Closes #167

## What was wrong

`provideCodeLenses` re-derived import detection line by line with `ImportFormatter.isModuleImport`, so it disagreed with `scanModuleImports` about what counts as an import. A `using` inside a `<# ... #>` block comment, or one indented inside a `module:` body, still got a **Use relative path** / **Use absolute path** lens and still counted towards the convert-all lens. Acting on either edited a line inside a comment.

`ImportPathConverter.convertAllImportsFromFullPath` and `convertAllImportsInDocument` had the same per-line shape and the same result, so **Use relative paths for all** converted those lines too.

## The change

`ImportScanner` gains `scanConvertibleImports(lines)`: the import statements a path conversion may act on. It is the full scan narrowed three ways, each for a reason conversion cares about:

- from `scanModuleImports`, so a `using` in a block comment or a module body is already gone;
- minus imports whose own line opens a block comment, since rebuilding such a line drops the opener and resurrects the region below it;
- minus the indented `using:` style, since every conversion path identifies an import by one line of statement text and replaces that one line.

The provider and both converter enumerations now read it. `rewritableImports` moves from `ImportDocumentEditor` to `ImportScanner` and is exported, since three callers need it now; that move is behaviour-neutral.

One behaviour change follows from the provider and the scanner agreeing, and is not a defect: `scanModuleImports` classifies with `{ atFileScope: true }`, so a file-level folder import written as a bare identifier — `using { Features }` — now gets a lens where the per-line detection read it as a local-scope using and offered none. It is a real module import at column 0, per the scanner's documented semantics. Both levels have a test, and the changelog fragment says so.

## Deliberately not in scope

`ImportPathConverter.applyConversion` still finds the line to edit by scanning for the first line whose trimmed text equals the statement, so with a commented-out copy of an identical import above the live one it can still edit the copy. Different trigger, and the fix is to thread through the line number the lens already passes and the command handlers already discard. Worth its own ticket.

## Tests

`src/imports/__tests__/ImportCodeLensProvider.test.ts` is new; 4 of its 9 cases fail against the old provider (block comment, module body, own-line opener, convert-all count). It needed `CodeLens`, `EventEmitter` and `workspace.onDidChangeTextDocument` in the `vscode` mock. `scanConvertibleImports` gets its own unit tests in `ImportScanner.test.ts`.

## Verification

```
$ npm run compile

> tsc -p ./
```

```
$ npm test

> jest --verbose

Test Suites: 18 passed, 18 total
Tests:       286 passed, 286 total
Snapshots:   0 total
Time:        6.24 s, estimated 7 s
Ran all test suites.
```

```
$ npm run format:check

> prettier --check "**/*.ts"

Checking formatting...
```

Not smoke-tested in the Extension Development Host.

## Review

Reviewed by a fresh model in a separate context: **PASS_WITH_SUGGESTIONS**, no Critical findings. All three findings were applied — the changelog no longer claims a guarantee `applyConversion` does not give, both convert-all counts are hoisted out of the loop as file-level properties, and the bare-identifier case is pinned by a test at both levels.


vukefn:
Filed the out-of-scope follow-up named above as #183: `applyConversion` still lands on the first line whose trimmed text matches the statement, so a commented-out copy of an identical import above the live one gets edited instead.

PR #184, opened 2026-08-06
fix(imports): land a path conversion on the line the lens was on

### LINKED ISSUE #141: Path conversion edits the first textually-matching import line, ignoring the line the user clicked
Found in a codebase audit after v0.8.0. Verified by executing the conversion against the real modules.

`ImportPathConverter.ts:531-545` finds its target with `lines[i].trim() === conversion.originalImport.trim()` and takes the first hit. The clicked line index is available — `ImportCodeLensProvider.ts:169,203` passes `i` — but `CommandsHandler.ts:406,474` names the parameter `_lineNumber` and drops it.

Duplicating an import at file scope and inside a `module:` body is idiomatic Verse, since a module body needs its own `using`:

```verse
using { Gadgets.Tools }          <- line 0

M := module:
    using { Gadgets.Tools }      <- line 3, the CodeLens the user clicked
```

Clicking the lens on line 3 produces the edit `replace (0,0)-(0,23)`. The wrong line is rewritten to the absolute path and the clicked line is untouched.

It self-heals on a second click, since line 0 no longer matches — but the first click silently converted an import the user did not select.

Fix: thread the line number through to `applyConversion` and use it to disambiguate, falling back to the text match only when the line does not match.

### LINKED ISSUE #183: applyConversion edits the first line matching the statement text, not the line the lens was on
Found by the review gate on #167.

`applyConversion` locates the line to rewrite by scanning from the top of the document for the first line whose trimmed text equals the statement it was given (`src/imports/ImportPathConverter.ts:534-540` on `main`; #182 shifts these by a few lines):

```ts
let lineIndex = -1;
for (let i = 0; i < lines.length; i++) {
    if (lines[i].trim() === conversion.originalImport.trim()) {
        lineIndex = i;
        break;
    }
}
```

Statement text is not unique in a document, so the first match is not necessarily the line the user acted on. Where an identical `using` sits inside a `<# ... #>` block comment above the live one, the conversion edits the commented copy:

```verse
<#
using { Gadgets.Tools }
#>

using { Gadgets.Tools }

code()
```

After #167 the lens correctly appears only on the live import at the bottom. Clicking **Use absolute path** on it rewrites line 1, inside the comment, and leaves the live import untouched. The same applies to a duplicate that is merely repeated rather than commented out: the second one is unreachable, and acting on it edits the first.

#167 fixed which lines *offer* a conversion; this is which line a conversion *lands on*, so it survives that fix.

The plumbing for the fix already exists and is thrown away. `ImportCodeLensProvider` passes the line index as the third command argument (`arguments: [document, trimmedLine, i]`), and both command handlers accept it and discard it — `CommandsHandler.convertToFullPath` and `convertToRelativePath` name the parameter `_lineNumber`. Threading it into `ImportConversionResult` (or into `applyConversion` alongside the conversion) and preferring it over the text search would make the edit land where the user clicked, with the search kept only as the fallback for the callers that have no line to offer.

Those callers are `convertAllImportsFromFullPath` and `convertAllImportsInDocument`, which enumerate through `scanConvertibleImports` after #182 and therefore already know each import's line — so they can supply it too, and the text search could go away entirely.

Worth checking in the same pass: `applyConversion` splits on `"\n"` while the document may be CRLF, so `lines[i].trim()` is doing double duty as both a whitespace trim and a `\r` strip. That is correct today but easy to break.

vukefn:
Closes #183
Closes #141

`applyConversion` found the line to rewrite by scanning from the top of the document for the first line whose trimmed text equalled the statement it was given. Statement text is not unique, so an identical `using` sitting in a `<# ... #>` block comment or plainly repeated above the clicked one absorbed the edit, and the line the user acted on was left untouched. #167 fixed which lines *offer* a conversion; this is which line one *lands on*.

#141 is the same defect reported from a `module:` body reproduction, so it closes with this.

## What changed

The line index was already threaded from `ImportCodeLensProvider` to the command handlers and dropped there as `_lineNumber`. It now travels on `ImportConversionResult` to `applyConversion`:

- `convertToFullPath` and `convertFromFullPath` take an optional line and record it on the result.
- `convertAllImportsInDocument` and `convertAllImportsFromFullPath` pass the line `scanConvertibleImports` already gives them, so the batch paths disambiguate too. A conversion replaces one line with one line, so an earlier apply never shifts a later recorded line.
- `applyConversion` resolves its target through `findConversionLine`, which takes the recorded line when that line still holds the statement, and otherwise searches.

The search is still needed: resolving a conversion spans a workspace scan, and an ambiguous one also spans a quick pick, so the document can change under a recorded line. It now searches the convertible imports the shared scanner reports rather than every line in the file — searching raw lines is what let the commented-out copy absorb the edit, and a fallback must not be able to land where a lens would never have appeared.

The CRLF concern the issue raises is handled: both sides of the comparison are trimmed, which absorbs the trailing `\r` a `"\n"` split leaves. `findConversionLine` now says so, since it was load-bearing and undocumented.

## Tests

Seven cases over `applyConversion` in `src/imports/__tests__/ImportPathConverter.test.ts`: the commented-out duplicate, two identical imports, indentation preserved on the named line, the two fallback paths, a fallback that must refuse a commented-out copy, and a statement absent from the document.

Proven to detect the defect. Against the parent commit, the three cases pinning the recorded line fail:

```
● ImportPathConverter.applyConversion › edits the line the conversion names, not an identical statement commented out above it
● ImportPathConverter.applyConversion › edits the named one of two identical imports rather than the first
Tests: 3 failed, 11 passed, 14 total
```

With the scanner-aware fallback removed, the sixth fails:

```
● ImportPathConverter.applyConversion › does not fall back onto a commented-out copy when the named line has moved
    Expected: false
    Received: true
Tests: 1 failed, 13 passed, 14 total
```

## Verification

```
> tsc -p ./

> jest --verbose

Tests:       303 passed, 303 total
Snapshots:   0 total
Time:        6.328 s
Ran all test suites.

> prettier --check "**/*.ts"

Checking formatting...
```

## Review gate

Two rounds with a fresh reviewer, no Critical in either.

Round 1, `PASS_WITH_SUGGESTIONS`: the changelog fragment was uncommitted, and the stale-line fallback was still the raw top-down search, so it could still land inside a commented-out copy. Both fixed in `22153f6`.

Round 2, `PASS_WITH_SUGGESTIONS`: the reviewer enumerated what the scanner-aware fallback now refuses that the old one accepted — copies inside `<# ... #>` comments, indented module-scoped copies, and lines whose text would not have matched anyway — and confirmed no legitimate edit is lost, since every reachable `originalImport` comes from `scanConvertibleImports` to begin with. Its remaining suggestion, restoring the indentation assertion in a reachable form, is `69fb940`.

## Noted, not changed

- `CommandsHandler.convertToFullPath` and `convertToRelativePath` ignore `applyConversion`'s return value and show a success status-bar message either way, so a refused edit still reads as "Using absolute path: ...". Pre-existing — the old `-1` path had it too — and the same class #180 fixed for `addSingleImport` and `optimizeImports`, left unfixed here. Worth a follow-up issue.
- `applyConversion` builds its replacement range as `[0, lines[i].length]` over a `"\n"` split, so on a CRLF document the range covers the trailing `\r` and the replacement drops it, leaving that one line LF-terminated. Pre-existing, separate from line selection, and out of scope here.
- `CommandsHandler` keeps a hand-mirrored `PathConversionResult` alongside the converter's own `ImportConversionResult`, with casts between them. Exporting the real interface would remove both, but that is a wider change than this ticket.
- The first commit on this branch carries a stray `@` in its subject, from a here-string that was written in the wrong shell dialect. The squash merge sets its subject explicitly, so it does not reach `main`.

### DIFF
--- a/changelog.d/183.fixed.md
+++ b/changelog.d/183.fixed.md
@@ -1,0 +1,1 @@
+**Path conversion**: Using a path conversion on an import whose statement also appears elsewhere in the file - commented out above it, say - now rewrites the line the CodeLens sits on instead of the first line reading the same.
--- a/src/commands/CommandsHandler.ts
+++ b/src/commands/CommandsHandler.ts
@@ -25,6 +25,7 @@ interface PathConversionResult {
     moduleName: string;
     isAmbiguous: boolean;
     possiblePaths?: string[];
+    line?: number;
 }
 
 /**
@@ -421,11 +422,13 @@ export class CommandsHandler {
     /**
      * Converts a single import to absolute path format.
      */
-    async convertToFullPath(document: vscode.TextDocument, importStatement: string, _lineNumber: number): Promise<void> {
+    async convertToFullPath(document: vscode.TextDocument, importStatement: string, lineNumber: number): Promise<void> {
         const documentUri = document.uri.toString();
         this.prepareForConversion(documentUri);
 
-        const result = (await this.deps.importPathConverter.convertToFullPath(importStatement, document.uri)) as PathConversionResult | null;
+        // The line the lens was clicked on travels with the conversion, so the
+        // edit lands there rather than on the first line reading the same.
+        const result = (await this.deps.importPathConverter.convertToFullPath(importStatement, document.uri, lineNumber)) as PathConversionResult | null;
 
         if (!result) {
             vscode.window.showInformationMessage("Import is already in absolute path format or could not be converted.");
@@ -489,11 +492,11 @@ export class CommandsHandler {
     /**
      * Converts a single import to relative path format.
      */
-    async convertToRelativePath(document: vscode.TextDocument, importStatement: string, _lineNumber: number): Promise<void> {
+    async convertToRelativePath(document: vscode.TextDocument, importStatement: string, lineNumber: number): Promise<void> {
         const documentUri = document.uri.toString();
         this.prepareForConversion(documentUri);
 
-        const result = (await this.deps.importPathConverter.convertFromFullPath(importStatement)) as PathConversionResult | null;
+        const result = (await this.deps.importPathConverter.convertFromFullPath(importStatement, lineNumber)) as PathConversionResult | null;
 
         if (!result) {
             vscode.window.showInformationMessage("Import cannot be converted to relative path.");
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -534,9 +534,13 @@ export class ImportPathConverter {
      * match is then the wrong line. Only that recorded line is checked against
      * the statement, never searched for, so a duplicate above it cannot win.
      *
-     * The text search remains for a caller with no line to give, and for a line
-     * that no longer holds the statement because the document changed while the
-     * conversion was being resolved.
+     * A search remains for a caller with no line to give, and for a line that no
+     * longer holds the statement because the document changed while the
+     * conversion was being resolved - resolving one spans a workspace scan, and
+     * an ambiguous one also spans a quick pick. That search goes through the
+     * shared scanner rather than every line, so a fallback cannot land somewhere
+     * a lens would never have appeared: searching the raw lines is what let the
+     * commented-out copy absorb the edit in the first place.
      *
      * Comparison is by trimmed text, which also absorbs the trailing `\r` a CRLF
      * document leaves on every element of a `"\n"` split.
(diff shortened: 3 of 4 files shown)


vukefn:
Filed the adjacent defect this branch does not touch: #185 - the conversion commands report success even when `applyConversion` refused the edit. Same class as #180, which covered `addSingleImport` and `optimizeImports` only.

PR #188, opened 2026-08-06
fix(imports): keep comments with the code they describe when organizing

### LINKED ISSUE #140: Optimize Imports moves imports above file header comments and orphans comments attached to imports
Found in a codebase audit after v0.8.0. Verified by executing `buildOrganizedContent` against the real modules.

`ImportDocumentEditor.ts:299-321` removes only the import lines and emits `[...formatted, "", ...remainingBody]`, so any comment line above or between imports is pushed below the regenerated block.

```
in : "# Copyright 2026 MyGame\n# Licensed MIT\n\nusing { /Verse.org/Simulation }\n\ncode()"
out: "using { /Verse.org/Simulation }\n\n# Copyright 2026 MyGame\n# Licensed MIT\n\n\ncode()"
```

```
in : "# Devices for the shop\nusing { /Fortnite.com/Devices }\n\ncode()"
out: "using { /Fortnite.com/Devices }\n\n# Devices for the shop\n\ncode()"
```

A license or attribution header ends up mid-file, and an explanatory comment is separated from the import it explains.

Note this is the organize path specifically. #90 fixed the analogous problem on the quick-fix path by preserving import-block location under `preserveImportLocations`; `buildOrganizedContent` does not honour the same invariant.

Related and lower severity: with `sortImportsAlphabetically: false` and an import block that does not start at line 0, `addImportsToDocument:237` inserts the new import above the header comment instead of merging it into the existing block, leaving two disconnected import regions.

vukefn:
Closes #140

## What was wrong

`buildOrganizedContent` removed the import lines and emitted `[...formatted, "", ...remainingBody]`, so every comment in the file landed below the rebuilt block. A licence or attribution header ended up in the middle of the file, and a comment written above an import was torn off the statement it explained. Both repro pairs in the issue reproduce exactly.

`addImportsToDocument` had the same defect in smaller form: with sorting off it inserted at line 0 whenever the first import block did not start there, which is precisely the file that opens with a header comment.

## What changed

**Two rules now decide where a comment goes.**

- The file's opening run of comment and blank lines is the header. It stays above the rebuilt block. A run of nothing but blank lines is not a header and is dropped as before.
- A run of comment lines written directly above an import is that import's annotation and travels with it, wherever the sort puts it. A blank line ends the run.

Line 0 is genuinely ambiguous - a file header and an annotation of the first import are written identically - and the header reading wins there. That is a deliberate call, documented on `headerLineCount` and pinned by a test: the alternative carries a licence into the middle of the import block when the import under it sorts away, which is the defect this change exists to remove.

**Comment structure comes from `ImportScanner`,** which already owns Verse's comment lexing. Deriving it in the editor from "does the line start with `#`" would attach a bare `#>` to an import and split a block comment in half. `classifyLines` reports `code` / `comment` / `blank` per line, plus whether the comment a line belongs to was opened above it, which is what stops a half-formed comment run from being carried away.

That includes the indented `<#>` comment: its body is every line below the marker indented past it, blank lines included (`00_overview.md:653` in the Verse book; the TextMate grammar's `end` pattern excludes `\s*$`). Block-comment depth still advances across that body exactly as before, so `blockCommentMask` and `scanModuleImports` see precisely what they saw before - there is a test asserting that.

**`addImportsToDocument`** now appends after the last import block wherever that block sits, and guards the document end: the line after the block does not exist in a file with no trailing newline, and VS Code clamps a position past the end onto the end of the last line, which spliced the new statement onto the last import and lost it.

## Deliberately out of scope

Reported rather than folded in, to keep the diff reviewable:

- The `preserveImportLocations: false` consolidation branch also inserts at position 0, above a header. The issue scopes itself to the organize path plus its one secondary note.
- Rebuilding an import line from its path alone still drops whatever trails it. `using { /A } # network only` comes back as `using { /A }`, which loses text; `using { /A } <#> why` comes back the same way, which also strands the marker's indented body as lines that no longer read as comment. Both are pre-existing, unchanged by this diff, and the same root cause - one issue, not this one. `rewritableImports` already excludes an import whose line opens a `<# ... #>` block for exactly this reason, so the `<#>` case is a small extension of a guard that exists.

## One existing expectation changed

`leaves a commented-out import in the body instead of resurrecting it` asserted that a `<# ... #>` region above the imports moves below them. Its intent - the commented-out import stays inert - is unchanged and still asserted; the region now stays where it was written, which is the whole point of the fix.

## Review

Reviewed by a fresh reviewer in a separate context over two rounds.

Round 1 returned **FAIL** on a real Critical: the first version of the classifier read a `<#>` marker's indented body as code, so the rebuilt block landed between a `<#>` header and its body, emptying the comment and stranding its text as indented lines at file scope - corruption rather than rearrangement. Confirmed against the language reference and fixed. Two Important findings were also fixed (the document-end insert clamp, and a leading blank run kept as a header). One Important - reading a line-0 comment as an annotation rather than a header - was declined with the reasoning above.

Round 2 returned **PASS_WITH_SUGGESTIONS** and withdrew that round-1 design finding, having reached the same conclusion independently: "the two readings of line 0 fail in different kinds rather than in different amounts". It verified the `<#>` fix by fuzzing - 7352 rebuilds over markers, indented bodies, tabs, block comments and blank lines in LF and CRLF with zero non-idempotent results and zero lost or duplicated comment lines, and 11811 cases asserting no line changes its lexical kind across a rewrite. The only remaining finding is the pre-existing trailing-`<#>` case listed under Deliberately out of scope, and the changelog sentence it flagged as over-claiming, which is now narrowed.

## Verification

`npm run compile`

```
> tsc -p ./

```

`npm test`

```
> jest --verbose

Tests:       328 passed, 328 total
Snapshots:   0 total
Time:        5.115 s, estimated 6 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

Every new regression test was proven to detect the defect: with the editor reverted and the tests kept, 11 of them fail (10 new, plus the one deliberately changed expectation), and they pass with the fix in place.


vukefn:
The two trailing-text cases listed under Deliberately out of scope are now tracked as #189. Both predate this PR and are unchanged by it; #189 also notes that the `opensIndentedComment` signal added here is what the `<#>` half of it would build on.

PR #190, opened 2026-08-07
fix(imports): keep the text trailing an import line through a rebuild

### LINKED ISSUE #189: Rebuilding an import line drops the text trailing it, deleting a comment or a `<#>` marker
Every writer in `ImportDocumentEditor` rebuilds an import line from its path alone, so whatever trails the statement on that line is discarded. Two cases, one root cause.

**Trailing line comment - text loss.**

```
in : "using { /Verse.org/Simulation } # network only\ncode()"
out: "using { /Verse.org/Simulation }\n\ncode()"
```

The annotation is gone with nothing to recover it from.

**Trailing `<#>` marker - the file stops parsing.** A `<#>` opens an indented comment whose body is every line below it indented past it (`00_overview.md:653` in the Verse book). Dropping the marker leaves that body as indented lines at file scope:

```
in : "using { /Verse.org/Simulation } <#> why this is here\n    it brings the simulation module into scope\ncode()"
out: "using { /Verse.org/Simulation }\n\n    it brings the simulation module into scope\ncode()"
```

## Where it comes from

`ImportFormatter.formatImportStatement` builds the line from the path, and `ScannedImport` carries no trailing text for a writer to put back. `rewritableImports` already refuses to rebuild an import whose line opens a `<# ... #>` block, for exactly this reason - see the doc comment on `ScannedImport.opensBlockComment`, which says the opener "lives in the trailing text a rebuild discards".

The `<#>` case is the same hazard with a different marker, and the signal for it now exists: `scanLine` in `ImportScanner` reports `opensIndentedComment` (added in #188). Folding that into `opensBlockComment`, or adding a sibling flag `rewritableImports` also filters on, would leave such a line alone the way a `<#` line already is.

The trailing line comment is a separate decision: excluding those imports from rewriting would stop them being sorted at all, so preserving the trailing text through the rebuild is probably the wanted behaviour rather than refusing to touch the line.

## Not a regression

Both predate #188 and are unchanged by it - the scanner returned early at a `<#>` before that change too, so the line was rebuilt identically. #188's reviewer surfaced them while fuzzing that PR (11811 cases asserting no line changes its lexical kind across a rewrite); these were the only cause of a kind change other than the documented `using:` pair normalization.

## Suggested acceptance

- An import line ending in a `<#>` marker is left where it is, like one opening a `<# ... #>` block, with its indented body intact.
- A trailing line comment survives a rebuild, in all three import styles.

vukefn:
Closes #189

Every writer rebuilt an import line from its path alone, so whatever trailed the statement on that line was discarded. Two symptoms, one root cause: `ScannedImport` carried nothing for a writer to put back.

## What changed

**A trailing comment now survives a rebuild.** `ScannedImport.trailingComment` holds it, extracted by a new `ImportFormatter.extractTrailingComment` that shares its split point with the existing `stripTrailingComment` (one private `commentStartIndex`, so the two halves cannot drift apart and always recompose the original). The four rebuild sites re-attach it, keyed on the formatted statement — the same handle `buildOrganizedContent` already uses to re-attach the comments written *above* an import, and the only one still available after a list of paths has been sorted and grouped.

**A line carrying a comment marker is left where it was written.** `opensBlockComment` becomes `anchorsCommentBelow` and covers the `<#>` indented-comment marker as well as the `<#` block opener it already covered. Renamed rather than extended in place: every consumer wanted the union and nothing told the two apart, and a field named "block" over a marker that the file's own doc comment insists is *not* a block opener would have been wrong.

The flag is read from the statement's own text — a `<#>` anywhere on its span, or a `<#` the span never closes — rather than from what happens to sit under it.

## The `<#>` semantics

The Book of Verse prose (`00_overview.md:653`) says a `<#>` sits "on its own line", which would make the trailing case in the ticket dubious. The compiler's own round-trip tests overrule it: `Tests/Roundtrip/Comments.versetest:309-320` asserts `a<#>` with `b<#>` indented below it round-trips, and `:227` has `<#>hey` mid-body. A `<#>` after code does open an indented comment.

## A defect the review caught

The first commit kept the pre-existing carve-out that left an unclosed `<#` on the buffer's last line rewritable, on the reasoning that it had no line below to swallow. That held only while a rebuild *discarded* the opener. Carrying it through and sorting it upward made it swallow the imports below:

```
in : "using { /Zebra }\nusing { /Apple } <#"
out: "using { /Apple } <#\nusing { /Zebra }\n"     <- /Zebra is now comment text
```

and through `addImportsToDocument` the newly added import landed under the opener, leaving it inert so auto-import re-fired on it forever. Both were reproduced as failing tests before the fix. Removing the carve-out also drops the wart where the outcome turned on whether the file ended with a newline.

## Tests

Each new test was proven to detect its own defect: the fix was reverted one behaviour at a time and the failures were disjoint — six for the marker anchoring, six for the comment preservation, no others in either run.

Coverage spans `scanModuleImports`, `scanConvertibleImports`, `extractTrailingComment`, `buildOrganizedContent` (all three import styles, dedup, idempotence, CRLF) and `addImportsToDocument` (both the block-replacement and consolidate writers, which the ticket's acceptance reaches and which had no trailing-comment coverage).

## Verification

```
> tsc -p ./

> jest --verbose

Tests:       365 passed, 365 total
Snapshots:   0 total
Time:        9.73 s
Ran all test suites.

> prettier --check "**/*.ts"

Checking formatting...
```

## Review gate

Round 1 FAIL (1 Critical, 1 Important, 6 Suggestions) — the Critical above. Round 2 PASS_WITH_SUGGESTIONS after the fix; both remaining Suggestions were applied (the trailing-comment helper now refuses an anchored import itself rather than relying on all four callers to pass a filtered list, and the regression test for the Critical was pinned to the exact edit instead of a vacuous "no `<#` anywhere" assertion).

## Out of scope, worth filing separately

- `ImportPathConverter.applyConversion` replaces the whole line with `originalIndent + finalImport`, so a path conversion still deletes a trailing comment. A different writer from the ones the ticket names; `ScannedImport.trailingComment` now carries what a fix would need.
- `buildOrganizedContent`'s `anchoredPaths` filters only `additionalPaths`, so a path imported both by an anchored line and a live one is emitted twice. Pre-existing for `<#`, newly reachable for `<#>`.

### DIFF
--- a/changelog.d/189.fixed.md
+++ b/changelog.d/189.fixed.md
@@ -1,0 +1,1 @@
+**Import comments**: the comment trailing a `using` statement is no longer deleted when imports are sorted or organized, and an import line carrying a `<#>` indented-comment marker is left where it is instead of being rebuilt without it.
--- a/changelog.d/189.fixed.md
+++ b/changelog.d/189.fixed.md
@@ -1,1 +1,1 @@
-**Import comments**: the comment trailing a `using` statement is no longer deleted when imports are sorted or organized, and an import line carrying a `<#>` indented-comment marker is left where it is instead of being rebuilt without it.
+**Import comments**: the comment trailing a `using` statement is no longer deleted when imports are sorted or organized. An import line carrying a `<#>` indented-comment marker, or an unclosed `<#` block opener, is now left where it was written rather than rebuilt without it, so the lines it comments out stay commented out.
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -142,7 +142,13 @@ export class ImportDocumentEditor {
     private trailingCommentsByStatement(scannedImports: ScannedImport[], preferDotSyntax: boolean): Map<string, string> {
         const byStatement = new Map<string, string>();
         for (const imp of scannedImports) {
-            if (!imp.trailingComment) {
+            // An anchored import has a trailing comment like any other, and its
+            // is the marker that anchors it. Restoring that onto a rebuilt line
+            // is what puts the marker somewhere it was not written, so it is
+            // refused here rather than only by every caller passing a filtered
+            // list. Every caller does today; one that forgot would resurrect a
+            // defect this branch has already shipped once.
+            if (!imp.trailingComment || imp.anchorsCommentBelow) {
                 continue;
             }
             const statement = this.formatter.formatImportStatement(imp.path, preferDotSyntax);
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -126,6 +126,48 @@ export class ImportDocumentEditor {
     }
 
     /**
+     * The comment trailing each import, keyed on the statement a rebuild emits
+     * for its path.
+     *
+     * Keyed on the formatted statement rather than on a line, because that is
+     * the only handle a writer still has once it has sorted and grouped a list
+     * of paths - the same reason buildOrganizedContent keys the comments
+     * written *above* an import that way.
+     *
+     * Deduplication keeps one statement for a repeated path, so the comments of
+     * its several occurrences are concatenated rather than replaced: keeping
+     * only the last would delete text the author wrote. They stay on one line,
+     * which is where they were, and a `#` comment runs to the end of it.
+     */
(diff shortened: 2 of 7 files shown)


vukefn:
Filed the two out-of-scope findings from the review: #191 (converter drops a trailing comment) and #192 (organize duplicates a path an anchored import provides). Both confirmed by running the code; #192 reproduces identically on `main`, so it predates this branch.

PR #203, opened 2026-08-08
fix(imports): keep the comment trailing a converted import line

### LINKED ISSUE #191: Converting an import path deletes the comment trailing its line
Converting an import path replaces the whole line, so a comment trailing that line is deleted.

`ImportPathConverter.applyConversion` builds the replacement from the statement alone:

```ts
const range = new vscode.Range(new vscode.Position(lineIndex, 0), new vscode.Position(lineIndex, lines[lineIndex].length));
const originalIndent = lines[lineIndex].match(/^\s*/)?.[0] || "";
edit.replace(document.uri, range, originalIndent + finalImport);
```

The range spans the entire line and the replacement carries only indent plus statement, so everything after the statement is dropped:

```
in : "using { Shop } # only the shop, not the vendor"
out: "using { /mygame@fortnite.com/mygame/Economy/Shop }"
```

## Why this is separate from #189

#189 fixed the same data loss in the writers named in its root-cause section - "every writer in `ImportDocumentEditor`". `ImportPathConverter` is a different writer and was out of that ticket's scope, so it was left alone deliberately rather than missed.

Structure is already safe here. `scanConvertibleImports` filters on `ScannedImport.anchorsCommentBelow`, so a line carrying a `<#` opener or a `<#>` marker gets no CodeLens and no conversion. What is still lost is annotation text only, which is the milder half of what #189 addressed.

## What a fix has to work with

`ScannedImport.trailingComment` (added by #189) already holds exactly the text a rebuild has to put back, and `ImportFormatter.extractTrailingComment` extracts it from a statement or a path line. `ConvertibleImport`, which is what the converter actually consumes, carries only `statement` and `line` - so either it grows the field, or `applyConversion` re-extracts from the line it is about to replace.

## Suggested acceptance

- Converting an import whose line ends in a `#` or `<# ... #>` comment leaves that comment on the converted line.
- The comment keeps its separation from the statement rather than being jammed against it.

Found by the review gate on #189 (PR #190), which confirmed the behaviour by reading the call site.

vukefn:
Closes #191

`ImportPathConverter.applyConversion` rebuilt the line from indent plus statement, so a comment trailing a `using` was deleted when its path was converted.

## What changed

**`applyConversion` restores the comment.** It is read with `ImportFormatter.extractTrailingComment` from the line the edit is about to overwrite, and joined with a single space — the same recomposition `ImportDocumentEditor.withTrailingComment` uses, so the two writers agree on separation.

The ticket offered a choice: grow `ConvertibleImport` with `trailingComment`, or re-extract in `applyConversion`. Re-extraction, because the comment then needs no field threaded through `ConvertibleImport` and `ImportConversionResult` to reach the one place that writes it.

**`convertFromFullPath` no longer bakes the comment into the path.** It held a third copy of the path regexes that skipped the trailing-comment strip. The dotted capture `/using\.\s*(.+)/` runs to end of line, so `using. /mygame@fortnite.com/mygame/Economy/Shop # see Economy/Vendor` produced the path `/…/Economy/Shop # see Economy/Vendor`, and the `split("/")` / `join(".")` emitted `using. Economy.Shop # see Economy.Vendor` — the `/` inside the comment rewritten as a `.`.

This is required by the ticket's acceptance rather than adjacent tidying: without it, restoring the comment would write it twice on every dotted conversion. It now reads through `extractPathFromImport` like its siblings. Behaviour is otherwise identical — same regexes, same curly-before-dot precedence, and the `null` vs `""` difference is absorbed by the guard on the next line.

## Scope

Structure was already safe and stays untouched: `scanConvertibleImports` excludes `anchorsCommentBelow`, so a line carrying an unclosed `<#` or a `<#>` marker never reaches the converter. Only annotation text was at stake.

Not fixed here, and filed separately: `usesCurlyBraces = originalImport.includes("{")` reads the raw statement including its comment at three sites, so a comment containing `{` flips the emitted syntax style. Pre-existing, a different symptom, and outside this ticket's acceptance.

## Tests

Six new cases in `src/imports/__tests__/ImportPathConverter.test.ts`, all without the VS Code runtime. Five fail against the unfixed converter — verified by stashing it: the braced `#` case, the `<# ... #>` inline block, indent-plus-comment, the dotted path staying clean, and the end-to-end no-duplication case. A CRLF case pins that the `\r` a `"\n"` split leaves behind does not ride into the restored comment. The braced-path case passes either way and pins the contract rather than catching the defect.

## Verification

```
$ npm run compile

> tsc -p ./

$ npm test

> jest --verbose

Tests:       402 passed, 402 total
Snapshots:   0 total
Time:        6.347 s, estimated 10 s
Ran all test suites.

$ npm run format:check

> prettier --check "**/*.ts"

Checking formatting...
```

Review gate: PASS_WITH_SUGGESTIONS, no Critical or Important findings. Two suggestions applied — the missing CRLF case, and a comment whose stated rationale described a `findConversionLine` divergence its matching rule cannot produce.

### DIFF
--- a/changelog.d/191.fixed.md
+++ b/changelog.d/191.fixed.md
@@ -1,0 +1,1 @@
+**Path conversion**: converting an import path keeps the comment trailing that line instead of deleting it, in both directions and in both the braced and dotted styles. A comment on a dotted absolute import is also no longer carried into the converted path, which previously rewrote any `/` inside it as a `.`.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -377,9 +377,13 @@ export class ImportPathConverter {
             return null;
         }
 
-        const curlyMatch = importStatement.match(/using\s*\{\s*([^}]+)\s*\}/);
-        const dotMatch = importStatement.match(/using\.\s*(.+)/);
-        const fullPath = curlyMatch ? curlyMatch[1].trim() : dotMatch ? dotMatch[1].trim() : null;
+        // Read through extractPathFromImport rather than a second copy of its
+        // regexes, because that copy skipped the trailing-comment strip: the
+        // dotted capture runs to end of line, so `using. /A/B # note` yielded the
+        // path `/A/B # note` and re-emitted the comment inside the converted
+        // statement - which applyConversion, restoring the comment itself, would
+        // then write a second time.
+        const fullPath = this.extractPathFromImport(importStatement);
 
         if (!fullPath || !fullPath.startsWith("/")) {
             return null;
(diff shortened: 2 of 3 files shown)


vukefn:
Filed the deferred style-detection defect as #204 — a `{` inside a trailing comment flips which `using` syntax a conversion emits, at three sites this PR deliberately left alone.

PR #206, opened 2026-08-08
fix(imports): judge conversion syntax style on the path, not the comment

### LINKED ISSUE #204: A brace inside a trailing comment flips the syntax style a path conversion emits
A comment containing `{` makes a path conversion switch the import to braced syntax.

Conversion decides which of the two `using` styles to emit by testing the raw statement for a brace:

```ts
const usesCurlyBraces = importStatement.includes("{");
```

The statement it tests is the whole trimmed line, comment included — `scanConvertibleImports` returns `lines[imp.startLine].trim()`. A `{` anywhere in the trailing comment therefore reads as a braced import:

```
in : "using. /mygame@fortnite.com/mygame/Economy/Shop # use {braces} here"
out: "using { Economy.Shop } # use {braces} here"
```

The path and the comment are both correct. Only the style flips, silently, against what the author wrote.

## Where

Three sites, all the same expression:

- `ImportPathConverter.convertFromFullPath` — `src/imports/ImportPathConverter.ts:413`
- `ImportPathConverter.convertToFullPath` — `src/imports/ImportPathConverter.ts:478`
- `ImportPathConverter.applyConversion`, the ambiguous branch — `src/imports/ImportPathConverter.ts:578`

## Root cause

Every other reader in the class routes through `extractPathFromImport`, which strips the trailing comment before looking at the content (`ImportPathConverter.ts:92-97`). The style test is the one place that reads the statement raw, so it is the one place a comment can still influence the result.

`ImportFormatter.stripTrailingComment` already accepts a whole statement — `using { A } # x` gives `using { A }` — so each site is a one-line change.

## Why this is separate from #191

#191 fixed the loss of the trailing comment itself, and made `convertFromFullPath` read its path comment-stripped. It deliberately left the style test alone: a flipped syntax style is a different symptom from a deleted comment, and nothing in #191's acceptance covers it. PR #203 does not make this better or worse at two of the three sites; at `convertFromFullPath` it turns an unparseable result into a merely wrong-styled one, because the comment no longer lands inside the braces.

## Suggested acceptance

- Converting an import whose trailing comment contains `{` keeps the style the author wrote, dotted or braced.
- The same holds for the ambiguous branch, where the user picks the path from a quick pick.

Found by the review gate on #191 (PR #203).

vukefn:
Closes #204

Path conversion picked between the dotted and braced `using` styles by testing the raw statement for a `{`. The statement is the whole trimmed line, so a brace anywhere in the trailing comment read as a braced import and the conversion silently rewrote the author's style:

```
in : using. /mygame@fortnite.com/mygame/Economy/Shop # use {braces} here
out: using { Economy.Shop } # use {braces} here
```

The style test was the one reader in `ImportPathConverter` that looked at the statement rather than at a capture taken from it, so it was the one place a comment could still reach. It now routes through `ImportFormatter.stripTrailingComment`, behind a single `usesBracedStyle` helper shared by the three sites that made the decision: `convertFromFullPath`, `convertToFullPath`, and the ambiguous branch of `applyConversion`.

## Tests

Five cases in `src/imports/__tests__/`, covering both halves of the rule at all three sites: a brace in the comment does not turn a dotted import braced, and stripping the comment does not cost a braced import its braces. The three brace-in-comment cases were each proven to fail against the unfixed code — stashing `ImportPathConverter.ts` gives 3 failed / 24 passed.

Edge cases checked and unaffected: inline `<# ... #>` (`commentStartIndex` backs up over the `<`), the indented `using:` style and comment-anchoring imports (both excluded by `scanConvertibleImports`), and local-scope `using{X}`.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
> jest --verbose

Test Suites: 21 passed, 21 total
Tests:       422 passed, 422 total
Snapshots:   0 total
Time:        12.078 s
Ran all test suites.

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

## Review gate

`PASS_WITH_SUGGESTIONS` — no Critical or Important findings. Three suggestions; the two that concerned this diff are applied in 8a3c4d0. The third, folding the triplicated `usesCurlyBraces ? ... : ...` emit ternary into `ImportFormatter.formatImportStatement`, is left alone as out of scope here.

The gate also surfaced a separate pre-existing defect, out of scope for this ticket and untouched by it: `extractPathFromImport` matches before it strips, so the curly regex can match inside a comment — `using. Economy.Shop # was using { Inventory }` extracts the path `Inventory`. The same ordering is in `ImportFormatter`.

### DIFF
--- a/changelog.d/204.fixed.md
+++ b/changelog.d/204.fixed.md
@@ -1,0 +1,1 @@
+**Path conversion**: a brace inside the comment trailing an import no longer flips the `using` syntax style the conversion emits. The style now follows the statement the author wrote, dotted or braced, with comment text excluded from the decision.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -96,6 +96,19 @@ export class ImportPathConverter {
         return ImportFormatter.stripTrailingComment(captured);
     }
 
+    /**
+     * Whether the author wrote the braced style rather than the dotted one.
+     *
+     * Judged on the statement minus any trailing comment: a `{` is legal in
+     * comment prose but not in a module path, so testing the raw line lets a
+     * comment decide the syntax the conversion emits. Every other reader in
+     * this class strips the comment first, via extractPathFromImport; this is
+     * the one that reads the statement rather than the path.
+     */
+    private static usesBracedStyle(importStatement: string): boolean {
+        return ImportFormatter.stripTrailingComment(importStatement).includes("{");
+    }
+
     /** Checks if an import is already in full path format */
     isFullPathImport(importStatement: string): boolean {
         const path = this.extractPathFromImport(importStatement);
@@ -410,7 +423,7 @@ export class ImportPathConverter {
             return null;
         }
 
-        const usesCurlyBraces = importStatement.includes("{");
+        const usesCurlyBraces = ImportPathConverter.usesBracedStyle(importStatement);
         const relativeImport = usesCurlyBraces ? `using { ${relativeImportPath} }` : `using. ${relativeImportPath}`;
 
         return {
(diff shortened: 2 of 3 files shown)


vukefn:
The separate defect noted above is now tracked as #207: `extractPathFromImport` matches before it strips, and its curly pattern is unanchored, so a comment carrying `using { X }` wins over a dotted statement's own path. Untouched by this PR.

PR #213, opened 2026-08-08
fix(imports): read an import path from the statement, not from its comment

### LINKED ISSUE #207: A comment containing "using {" hijacks the path read out of a dotted import
A comment containing the literal text `using {` hijacks the path read out of a dotted import.

`extractPathFromImport` matches before it strips. The curly regex is tried first and is unanchored, so it is free to match inside the trailing comment of a dotted statement and win over the dotted capture. Only then is the comment stripped — from the wrong capture.

```
in : using. Economy.Shop # was using { Inventory }
out: "Inventory"                                     (expected "Economy.Shop")

in : using. /mygame@fortnite.com/mygame/Economy/Shop # see using { Vendor }
out: "Vendor"                                        (expected the full path)

in : using. Economy.Shop # plain note
out: "Economy.Shop"                                  (correct)
```

The trigger is narrower than a brace alone: the comment has to carry `using` followed by a braced group, which is exactly what a comment cross-referencing another import looks like.

## Where

Two copies of the same body, both curly-first and both unanchored:

- `ImportFormatter.extractPathFromImport` - `src/imports/ImportFormatter.ts:171-183`
- `ImportPathConverter.extractPathFromImport` - `src/imports/ImportPathConverter.ts:92-97`

`ImportFormatter.extractPathFromImport` feeds `isDigestImport` and the formatting and editing paths; `ImportPathConverter.extractPathFromImport` feeds `isFullPathImport`, `isBuiltinModule` and `extractModuleFromImport`, so a hijacked path can misclassify an import as full-path or built-in as well as convert the wrong module.

## Root cause

Ordering, and anchoring. `ImportFormatter.isModuleImport` (`ImportFormatter.ts:130-148`) already has this right and is unaffected: it anchors each pattern with `^` and tries the dotted style before the braced one, so a dotted statement is classified by its own head and never by its comment. The two `extractPathFromImport` bodies do neither.

Either fix works on its own:

- anchor both patterns with `^` on the trimmed statement, so a match inside a comment cannot start one; or
- strip the trailing comment from the statement before matching, rather than from the capture afterwards.

Anchoring is the closer match to what `isModuleImport` already does, and reversing the two branches to dotted-first would make the pair consistent.

## Suggested acceptance

- A dotted import whose trailing comment contains `using { X }` yields its own path, not `X`.
- The same for both copies of `extractPathFromImport`.

Pre-existing, and untouched by #206 - that PR fixed the syntax-style test, which is a different reader. Found by the review gate on #204 (PR #206), where the wrong invariant was briefly written into a docstring before being corrected.

vukefn:
Both copies of `extractPathFromImport` tried the braced pattern first and neither anchored it, so on a dotted statement the braced regex was free to match inside the trailing comment and win the path. The comment was only stripped afterwards, from that wrong capture.

```
in : using. Economy.Shop # was using { Inventory }
out: "Inventory"                      (expected "Economy.Shop")
```

Both now trim, anchor, and try dotted before braced, following `ImportFormatter.isModuleImport`, which already reads a dotted statement by its own head rather than by its comment.

## What that cost, and what it bought

Anchoring means a statement has to open the text it is read from. `allUsingPaths` deliberately required no such thing: a live `using` can follow the `#>` that closes a block comment, closed inline trivia, or a `;`. Three existing tests caught this immediately, and the review gate then caught a subtler miss in the first correction — reading from a single "where the code starts" index still lost `Foo(); using { LocalVar }`, which under the old unanchored search was reported. That is the one error the function must not make: its caller reads an empty answer as permission to withhold an import.

So `scanLine` now reports the line **with its comments removed** rather than where its code starts, and `allUsingPaths` offers each `using` in that text to `extractPathFromImport` in turn. The whole line is searched, as before; comment text simply is not part of what is searched, which is what the anchoring was for.

Nothing replaces a removed comment, so text on either side joins. Whether Verse splits a token there is not settled by its test corpus — comments sit between tokens throughout `Tests/Roundtrip/Comments.versetest` and never inside an identifier, and the one splicing case, `"abc<#def#>ghi" = "abcghi"`, is string contents rather than code. With the rule undecided the direction is what matters, and this reader has one safe direction: an invented path makes its caller more cautious, a missed one is read as permission to withhold an import.

Reading from where the code starts also stops a code line's trivia reporting a path — `code() # see using { /B }` used to report `/B`. The old doc comment called that over-reporting an acceptable cost; it is the same defect one level up, and it is now gone.

## Files

- `src/imports/ImportFormatter.ts` — anchored, dotted-first.
- `src/imports/ImportPathConverter.ts` — the same, keeping its `""`-on-no-match contract, which `isFullPathImport` and `isBuiltinModule` call `.startsWith` on unguarded.
- `src/imports/ImportScanner.ts` — `LineScan.code` / `LineClassification.code`, and `firstImportPath`.

`isDigestImport` improves as a side effect: it routes anything containing the substring `using` through the extractor, which a bare path can satisfy (`/mygame@fortnite.com/mygame/Housing.Shop` — "Ho**using**."). It used to return `Shop`; it now falls through to its `|| importPath` fallback and keeps the real path.

## Not done here

Two things, both pre-existing and both deliberately left:

- The two `extractPathFromImport` bodies remain two copies rather than one shared helper. Folding them together changes `ImportFormatter`'s public shape, which is wider than this fix; the ticket asks for the pair to be made consistent, and they now are.
- `allUsingPaths` still reports one path per line, so `using { /X }; using { Economy.Shop }` reports only `/X`. The unanchored search it replaced stopped at its first match too, so this is not a regression, but the function's doc promises the path of *every* `using`. Worth its own issue.

## Tests

Nine added, across `ImportFormatter.test.ts`, `ImportPathConverter.test.ts` and `ImportScanner.test.ts`. Six were proved against the unfixed sources — stash the fix, watch them fail, restore — and two of the three scanner guards were proved against the interim version the review rejected. The ninth pins the comment-splice direction, which is a choice rather than a defect, so it has no pre-fix failure to demonstrate.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
> jest --verbose

Test Suites: 22 passed, 22 total
Tests:       452 passed, 452 total
Snapshots:   0 total
Time:        5.781 s, estimated 7 s
Ran all test suites.

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

Reviewed in a fresh context over two rounds: round 1 `FAIL` on one Critical — the miss described above — plus three Suggestions; round 2 `PASS_WITH_SUGGESTIONS` after the fix, with the reviewer re-implementing the pre-fix `allUsingPaths` and diffing it against this branch across 21 line shapes to confirm no real path is lost.

Closes #207

### DIFF
--- a/changelog.d/207.fixed.md
+++ b/changelog.d/207.fixed.md
@@ -1,0 +1,1 @@
+**Import paths read from the statement, not its comment**: a `using` import whose trailing comment mentioned another import in braces, such as `using. Economy.Shop # was using { Inventory }`, had that other path read as its own - converting the wrong module and misclassifying the import as a full path or a built-in one.
--- a/src/imports/ImportFormatter.ts
+++ b/src/imports/ImportFormatter.ts
@@ -165,8 +165,8 @@ export class ImportFormatter {
      * A trailing comment is stripped, so the returned path never carries trivia
      * that would corrupt the statement when it is re-emitted.
      *
-     * Anchored on the trimmed statement, and dotted before braced, exactly as
-     * isModuleImport does it. Unanchored, the braced pattern is free to match
+     * Anchored on the trimmed statement, and dotted before braced, following
+     * isModuleImport. Unanchored, the braced pattern is free to match
      * inside the trailing comment of a dotted statement and win the path -
      * `using. Economy.Shop # was using { Inventory }` read as `Inventory`,
      * because the comment is only stripped afterwards, from that capture.
--- a/src/imports/ImportFormatter.ts
+++ b/src/imports/ImportFormatter.ts
@@ -164,19 +164,28 @@ export class ImportFormatter {
      * Handles both curly syntax (using { /path }) and dot syntax (using. /path).
      * A trailing comment is stripped, so the returned path never carries trivia
      * that would corrupt the statement when it is re-emitted.
+     *
+     * Anchored on the trimmed statement, and dotted before braced, exactly as
+     * isModuleImport does it. Unanchored, the braced pattern is free to match
+     * inside the trailing comment of a dotted statement and win the path -
+     * `using. Economy.Shop # was using { Inventory }` read as `Inventory`,
+     * because the comment is only stripped afterwards, from that capture.
+     *
      * @param importStatement The full import statement
      * @returns The extracted path, or null if there is none (including a
      *   statement whose content is nothing but a comment)
      */
     extractPathFromImport(importStatement: string): string | null {
-        const curlyMatch = importStatement.match(/using\s*\{\s*([^}]+)\s*\}/);
-        if (curlyMatch) {
-            return ImportFormatter.stripTrailingComment(curlyMatch[1]) || null;
-        }
+        const trimmed = importStatement.trim();
 
-        const dotMatch = importStatement.match(/using\.\s*(.+)/);
+        const dotMatch = trimmed.match(/^using\.\s*(.+)/);
         if (dotMatch) {
             return ImportFormatter.stripTrailingComment(dotMatch[1]) || null;
+        }
+
+        const curlyMatch = trimmed.match(/^using\s*\{\s*([^}]+)\s*\}/);
+        if (curlyMatch) {
+            return ImportFormatter.stripTrailingComment(curlyMatch[1]) || null;
         }
 
         return null;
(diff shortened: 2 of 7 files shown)


vukefn:
Follow-up filed as #214 — `allUsingPaths` reports one path per line rather than every `using`, which the review gate found while checking this branch. Pre-existing and unchanged here, so it stays out of this diff.

PR #215, opened 2026-08-08
fix(imports): report every using a line writes, not just its first

### LINKED ISSUE #214: allUsingPaths reports one path per line, not every using the file writes
`allUsingPaths` promises "the path of every `using` the file writes", but reports at most one per line. A line carrying two live statements yields only the first:

```
in : ["using { /X }; using { Economy.Shop }"]
out: ["/X"]                                   (expected ["/X", "Economy.Shop"])
```

## Why it matters

The single caller is `withholdIsSafe` (`src/imports/ImportDocumentEditor.ts:629`), which reads the answer to decide whether an import that an anchored one duplicates may be withheld. An all-absolute answer reads as safe. The combination that bites is a rank-0 path written before a rank-1 or rank-2 one on the same line, exactly as above: `/X` is seen, `Economy.Shop` is not, the judgement flips to "safe", and the file loses a provider a module-body `using` was resolving against.

That is the error the function's own doc comment says it must never make - "the caller reads an empty answer as permission to remove an import, so anything unrecognised has to fail towards reporting a path rather than away from it".

## Where

`firstImportPath` (`src/imports/ImportScanner.ts:384`) returns on the first `using` that yields a path:

```ts
for (let at = code.indexOf("using"); at !== -1; at = code.indexOf("using", at + 1)) {
    const path = formatter.extractPathFromImport(code.slice(at));
    if (path) {
        return path;
    }
}
```

The loop that would collect them all is already written; only the `return` is early.

## Not a regression

Pre-existing. The unanchored search this replaced stopped at its first match too, so the behaviour is unchanged by #207 - which is why #207 left it alone rather than widening its diff. Found by the review gate on #207 (PR #213), whose reviewer verified the pre-#207 behaviour by re-implementing it and diffing across 21 line shapes.

## Suggested acceptance

- Both paths of `["using { /X }; using { Economy.Shop }"]` are reported, in written order.
- The existing one-path-per-line tests in `src/imports/__tests__/ImportScanner.test.ts` keep passing, the indented `using:` pair included - it is counted once, from its path line.

vukefn:
Closes #214

`allUsingPaths` promised the path of every `using` the file writes but read at most one per line: the loop behind it returned on the first statement that yielded a path. A `;` separates definitions in a scope exactly as a newline does, so a line can carry two live statements, and a rank-0 path written before a rank-1 or rank-2 one hid the second from `withholdIsSafe` - whose all-absolute answer is what permits an import to be withheld. The stranded provider is a file that stops compiling.

The loop that collects them all was already written; only the `return` was early.

## What changed

- `firstImportPath` becomes `usingPathsOnLine`, returning every path the line writes in written order instead of the first. Named for its scope, since the file-level `allUsingPaths` sits beside it.
- `allUsingPaths` spreads that result. The indented `using:` pair keeps its own branch and is still counted once, from its path line.
- Five tests in `src/imports/__tests__/ImportScanner.test.ts`, which run without the VS Code runtime.

Nothing else moved. `extractPathFromImport` is untouched, so its four other callers are unaffected.

## The dotted statement

A dotted `using` has no closing delimiter, so one sharing a line swallows what follows it into its path: `using. /X; using { Economy.Shop }` reports `/X; using { Economy.Shop }` as well as `Economy.Shop`. The malformed entry is reported rather than suppressed, and the statement after it is found by the next iteration and reported on its own - which is the guarantee the function owes its caller. Bounding the dotted capture at `;` would mean changing `extractPathFromImport`, shared by four other call sites, to fix cosmetics in the one output that is never re-emitted: `allUsingPaths` has a single caller and it only ranks these paths.

The change is one-directional by construction. Every path the old code returned is still returned, in the same position, with more appended - so `.every(rank === 0)` can only go true to false, never false to true. No input flips a judgement towards "safe".

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Test Suites: 23 passed, 23 total
Tests:       470 passed, 470 total
Snapshots:   0 total
Time:        9.053 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

The new tests were proved to detect the defect: stashing only `src/imports/ImportScanner.ts` and re-running the file fails 4 of the 5. The fifth, "does not read a second statement out of a comment sharing the line", is a guard that must pass both ways.

## Known hole, not this ticket

An indented pair opened after a `;` still under-reports: `["using { /X }; using:", "    Economy.Shop"]` returns `["/X"]`, verified by running it. The `using:` branch tests the whole trimmed line, so the shape misses it. Behaviour is identical before and after this diff and outside the ticket's acceptance, so it is left for a separate issue rather than widening this one.

### DIFF
--- a/changelog.d/214.fixed.md
+++ b/changelog.d/214.fixed.md
@@ -1,0 +1,1 @@
+**Optimize Imports**: a file that writes two `using` statements on one line no longer loses an import the code below it needs.
--- a/changelog.d/214.fixed.md
+++ b/changelog.d/214.fixed.md
@@ -1,1 +1,1 @@
-**Optimize Imports**: a file that writes two `using` statements on one line no longer loses an import the code below it needs.
+**Optimize Imports**: a file holding both a comment-anchored import and an ordinary import of the same path no longer removes the ordinary one when two `using` statements share a line and something below still needs one of them.
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -375,23 +375,36 @@ export function rewritableImports(scannedImports: ScannedImport[]): ScannedImpor
 }
 
 /**
- * The path of the first `using` written anywhere in a line of live code, or ""
- * when it writes none.
+ * The path of every `using` written anywhere in a line of live code, in written
+ * order, or [] when it writes none.
  *
  * extractPathFromImport reads a statement from the head of what it is given, so
  * a statement that does not open the line has to be handed its own head rather
  * than the line. Every `using` on the line is offered in turn, which keeps one
  * copy of the two patterns instead of a looser second copy that searches - the
  * looser copy is what read a comment as the statement in the first place.
+ *
+ * All of them are collected rather than the first, because a line can carry more
+ * than one live statement - `;` separates definitions in a scope exactly as a
+ * newline does. Stopping at the first is how a rank-0 path written before a
+ * rank-1 or rank-2 one on the same line hid it from the caller, which reads an
+ * all-absolute answer as permission to remove an import.
+ *
+ * A dotted statement has no closing delimiter, so one sharing a line swallows
+ * what follows it into its path. The next iteration still finds the statement
+ * after it and reports that path on its own, so nothing is missed - which is the
+ * guarantee this owes its caller. The swallowed copy only adds a path the caller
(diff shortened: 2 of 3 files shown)


vukefn:
The remaining hole named under "Known hole, not this ticket" is now tracked as #216.

PR #218, opened 2026-08-08
fix(imports): read a using: pair opened after a semicolon

### LINKED ISSUE #216: allUsingPaths misses an indented using: pair opened after a semicolon
`allUsingPaths` reads an indented `using:` pair only when the pair opens its own line, so a pair written after a `;` is missed along with the path on the line below it:

```
in : ["using { /X }; using:", "    Economy.Shop"]
out: ["/X"]                                   (expected ["/X", "Economy.Shop"])
```

Verified by running it against `main` at f3dba89 and against #215.

## Why it matters

The same reason #214 mattered, and the same caller. `withholdIsSafe` (`src/imports/ImportDocumentEditor.ts:629`) reads an all-absolute answer as permission to withhold an import that an anchored one duplicates. Here `/X` is rank 0 and `Economy.Shop` is never reported, so the judgement flips to "safe" and the file loses a provider the pair was resolving against.

## Where

`allUsingPaths` (`src/imports/ImportScanner.ts:466`) decides the shape with a whole-line test:

```ts
if (/^using\s*:\s*$/.test(trimmed)) {
```

`using { /X }; using:` does not match it, so the line falls through to `usingPathsOnLine`, which reads statements from the line's own text. That helper finds `/X` from the first statement and nothing from the `using:` fragment, which matches neither the braced nor the dotted pattern. The path line below holds no `using` of its own, so it contributes nothing either - by design, since the pair is meant to be counted once from the line that opens it.

## Not a regression

Pre-existing, and unchanged by #215: `firstImportPath` returned `/X` for that line before, and `usingPathsOnLine` returns `["/X"]` now. #215 fixed the several-statements-per-line case and deliberately left this one, because it is the one shape whose second half lives on the next line.

## The open question

Whether `using:` opening an indented block after a `;` on the same line is legal Verse at all. The Book of Verse constrains where a `using` may appear, not that it must own its line, and the compiler's own tests write `using{...}; <definition>` at module scope - but neither settles a block-opening `using:` mid-line. If it is not legal, the fix is smaller: the shape only has to fail towards reporting rather than parse correctly.

## Suggested acceptance

- `["using { /X }; using:", "    Economy.Shop"]` reports both paths, or reports something that ranks non-zero, so `withholdIsSafe` cannot read it as safe.
- The existing `allUsingPaths` tests keep passing, the plain indented pair included - still counted once, from its path line.

vukefn:
`allUsingPaths` recognised the `using:` that opens an indented pair only when it was the whole of its line, so a pair written after a `;` lost the path on the line below it:

```
in : ["using { /X }; using:", "    Economy.Shop"]
out: ["/X"]                                   (expected ["/X", "Economy.Shop"])
```

The line then reported nothing but its own absolute statements, and `withholdIsSafe` reads an all-absolute answer as permission to withhold an import an anchored one duplicates - so the file lost a provider the pair was resolving against.

## The change

The opener is matched as the tail of the line rather than as the whole of it, and `usingPathsOnLine` now runs on every line unconditionally, reporting whatever precedes the opener.

A `using:` carries no path of its own - `extractPathFromImport` matches neither the dotted (`^using\.`) nor the braced (`^using\s*\{`) pattern against it - so running it through that helper first contributes nothing, and the pair is still counted once, from its path line.

Output is a superset of the old output for every input: `^using\s*:\s*$` is a strict subset of `\busing\s*:\s*$` on a trimmed string, and `usingPathsOnLine` now runs on a superset of lines. That direction matters, because the caller reads a missing path as permission to delete an import.

`\b` keeps the match off an identifier ending in the same letters (`Xusing:`, `My_using:`), which is pinned by a test.

## On the ticket's open question

Whether a block-opening `using:` after a `;` is legal Verse is left unanswered, and the fix does not depend on it. The Book of Verse treats `using` as a statement-level declaration and never writes the indented form; the compiler corpus attests `using{...}; <definition>` on one line and a block opener after a `;` mid-line separately, but not the combination. The fix errs towards reporting either way, which the ticket explicitly permits.

## Scope

`scanModuleImports` still decides the indented pair with the old whole-line test, so the two functions in this file now disagree about whether a `using:` may follow a `;`. That is pre-existing and out of scope here; noted for a follow-up rather than widened into this PR.

## Verification

`npm run compile`

```
> tsc -p ./

```

`npm test`

```
> jest --verbose

Test Suites: 25 passed, 25 total
Tests:       496 passed, 496 total
Snapshots:   0 total
Time:        9.311 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

The two behavioural regression tests were confirmed to fail against the unfixed scanner - stashing `ImportScanner.ts` and re-running the suite reproduced the issue's symptom exactly, `Economy.Shop` missing from both:

```
Tests:       2 failed, 77 passed, 79 total
```

Review gate: PASS_WITH_SUGGESTIONS, 0 Critical. The suggestions taken were comment accuracy and length in the hunk itself.

Closes #216

### DIFF
--- a/changelog.d/216.fixed.md
+++ b/changelog.d/216.fixed.md
@@ -1,0 +1,1 @@
+**Organize imports**: an indented `using:` pair written after a `;` on the same line no longer reads as absent, so organizing keeps an import the pair resolves against instead of dropping it.
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -444,8 +444,10 @@ function usingPathsOnLine(formatter: ImportFormatter, code: string): string[] {
  * A `;` also means a line can carry more than one statement, so a line
  * contributes every `using` statement usingPathsOnLine finds on it, in written
  * order, rather than the first. Reporting the first was the same error under
- * another name - see that function. The indented `using:` pair is the one shape
- * still counted once, because its path is read here from the line below.
+ * another name - see that function. For the same reason a `using:` opening an
+ * indented pair is recognised at the end of its line rather than as the whole
+ * of it: a statement can precede it there too. The pair is the one shape still
+ * counted once, because its path is read here from the line below.
  *
  * Positions are not reported; a caller that needs them wants scanModuleImports.
  */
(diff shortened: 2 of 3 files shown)


vukefn:
Follow-up filed as #219 for the scanModuleImports half noted under Scope: the same whole-line test in the writer rebuilds `using { /X }; using:` from `/X` alone, deleting the opener and stranding the path below it. Pre-existing - the organized output is byte-identical on origin/main and on this branch.

Issue #223, opened 2026-08-08
Organizing deletes the second import on a line carrying two using statements


vukefn:
`scanModuleImports` records one import per line, so a line carrying two braced statements is recorded as its first path alone. Organizing then rebuilds the line from that path and the second statement is gone:

```
in : using { /X }; using { /Y }

     code()

out: using { /X }

     code()
```

`/Y` is not imported any more. Unlike #219 nothing is left behind to signal the loss - the output is a well-formed import block that silently imports one path fewer than the input.

## Where

`scanModuleImports` (`src/imports/ImportScanner.ts:342`) ends its loop by reading a single path from the head of the line:

```ts
const path = formatter.extractPathFromImport(trimmed);
if (path) {
    imports.push({ path, startLine: i, endLine: i, ... });
}
```

`extractPathFromImport` reads the statement at the head of what it is given, so `/Y` is never seen. The recorded entry spans line `i` and is rewritable, so `hoistedLines` takes the line and the writer re-emits it as `using { /X }`.

```
scanModuleImports(["using { /X }; using { /Y }", "", "code()"])
  -> [{ path: "/X", startLine: 0, endLine: 0, anchorsCommentBelow: false, trailingComment: "" }]
```

`allUsingPaths` on the same input returns `["/X", "/Y"]` - #215 taught the reader that a line can carry more than one `using`, and that lesson was not carried into the writer.

## Relationship to #219

Same root assumption - that a recorded import accounts for its whole line - and the same family of damage. They are separate because the symptom and the fix differ:

- #219 is a `using:` pair opened after a `;`. The pair's path is stranded below as a bare indented line, so the damage is visible in the output.
- This is two complete statements on one line. The second simply disappears.

#219's fix pins the pair's span so no writer rebuilds it, and deliberately does not detect this shape - `ScannedImport.rebuildLosesText` is documented as "no loss known" rather than "provably none", and points here.

## Suggested acceptance

- Organizing `["using { /X }; using { /Y }", "", "code()"]` either keeps the line intact, or rebuilds it into a form that still imports both paths - never one that imports only the first.
- A line carrying one statement is unaffected, the indented pair included.
- A regression test in `src/imports/__tests__/`, which runs without the VS Code runtime.

## The open question

Whether the fix should pin such a line, as #219 does, or record every statement on it and let the writer split them into separate lines. Splitting is more useful and changes text the author chose to keep on one line; pinning is conservative and leaves the file untidy. Worth settling before implementing.

[...]

PR #231, opened 2026-08-08
fix(imports): pin a using: pair opened after a semicolon

### LINKED ISSUE #219: Organizing mangles a using: pair opened after a semicolon, stranding its path
`scanModuleImports` reads `using { /X }; using:` as a plain one-line braced import of `/X`, so organizing rebuilds that line from the path alone - deleting the `; using:` that opened the pair and stranding the path below it in the body:

```
in : using { /X }; using:
         Economy.Shop

     code()

out: using { /X }

         Economy.Shop

     code()
```

`Economy.Shop` is no longer an import. It is left as a bare indented line the pair used to introduce.

## Where

`scanModuleImports` (`src/imports/ImportScanner.ts:322`) decides the indented pair with a whole-line test:

```ts
if (/^using\s*:\s*$/.test(trimmed)) {
```

`using { /X }; using:` does not match, so the line falls through to `extractPathFromImport(trimmed)`, which reads `/X` from the first statement and records a single-line import:

```
scanModuleImports(["using { /X }; using:", "    Economy.Shop", "code()"])
  -> [{ path: "/X", startLine: 0, endLine: 0, anchorsCommentBelow: false, trailingComment: "" }]
```

`endLine === startLine`, so only line 0 enters `hoistedLines` and line 1 stays in the body while the text that gave it meaning is rewritten away.

## Relationship to #216 and #218

This is the same whole-line test, in the sibling function. #218 fixed it in `allUsingPaths` only, and deliberately did not widen into the writer - so the file now holds two functions that disagree about whether a `using:` may follow a `;`.

Pre-existing and unaffected by #218. Verified by running the repro against `origin/main` and against the merged #218 branch: the organized output is byte-identical, and only `allUsingPaths` differs (`["/X"]` before, `["/X", "Economy.Shop"]` after). `withholdIsSafe` is not involved - the input has no anchored import, so `anchoredPaths` is empty and the withhold filter is inert either way.

## Why it is worse than #216

#216 was a missed *signal* - the file lost an import it already had a duplicate of. This rewrites the text directly: the pair's opener is deleted and its path is left as a dangling expression, so the damage does not depend on any other import being present.

## The open question, unchanged from #216

Whether a block-opening `using:` after a `;` is legal Verse is still unsettled. The Book of Verse treats `using` as a statement-level declaration and never writes the indented form; the compiler corpus attests `using{...}; <definition>` on one line and a `:` opener following a `;` mid-line separately, but not the combination.

It matters less here than it did for #216, because a writer must not mangle text it does not understand whichever way the question falls. If the shape is illegal, the fix is to leave the line alone rather than to parse it.

## Suggested acceptance

- Organizing `["using { /X }; using:", "    Economy.Shop", "", "code()"]` either keeps both lines intact, or rebuilds them into a form that still imports both paths - never one that drops `; using:` while leaving the path line behind.
- The existing `scanModuleImports` tests keep passing, the plain indented pair included - still consumed as one entry spanning both lines.

Found during the #216 review gate.

vukefn:
Closes #219

`scanModuleImports` decided the indented pair with a whole-line test, so `using { /X }; using:` read as a plain one-line import of `/X`. Organizing rebuilt that line from the path alone - deleting the `; using:` that opened the pair and leaving `Economy.Shop` below it as a bare indented expression that imports nothing.

## The fix

The invariant the scanner owes every writer is that a recorded import's span is reproducible from its `path`, because that is all a writer rebuilds from. A pair opened after a `;` breaks it: the span says more than any one path can.

So the scanner now reads that shape, and records **every path such a line writes as pinned** - present for existence and deduplication, never rebuilt or moved. That reuses the mechanism `anchorsCommentBelow` already has: a new `ScannedImport.rebuildLosesText`, filtered by `rewritableImports` alongside it. `ImportDocumentEditor` keys its suppression set on the line staying put rather than on why it does, so a path a pinned line already provides is still never written a second time.

The tail is tested against the line's **code** rather than its raw text, as `allUsingPaths` already is. A `$` anchored on the raw line is defeated by any comment after the `using:`, which puts the line straight back on the mangling path; searching the raw text has the mirror failure, reading a `using:` written inside a comment as an opener and recording the comment text below it as an import path. Both were caught by the review gate, with reproductions.

This also makes the two functions agree, which is what #219 opens with: #218 fixed the same whole-line test in `allUsingPaths` and deliberately did not widen into the writer.

## Behaviour

| input | before | after |
| --- | --- | --- |
| `using { /X }; using:` + `    Economy.Shop` | `using { /X }`, path stranded below | both lines intact, both paths counted as imported |
| `using { /X }; using: # note` + path line | same corruption, plus `# note` | both lines intact |
| `using { /X }; using:` opening nothing usable | `using { /X }`, `; using:` deleted | line intact, `/X` still counted |
| `using:` + `    Economy.Shop` (plain pair) | one rewritable entry | unchanged |
| `X := 1; using:` + path line | skipped whole | unchanged |

## Not in this PR

- **#223**, filed from this work: `using { /X }; using { /Y }` organizes to `using { /X }`, deleting `/Y` outright with nothing left behind to signal the loss. Same root assumption, different symptom, no acceptance criterion in #219. `rebuildLosesText` is documented as "no loss known" rather than "provably none" and points at it.
- **A pre-existing ordering hole, not introduced here and not fixed here.** A pinned line stays where it is, so a consumer can still be hoisted above the provider it resolves against: `["using { Features }; using:", "    Economy.Shop", "using { Economy.Other }"]` puts `Economy.Other` on top, above the line bringing `Economy` into scope. The identical shape with an anchored import produces the identical ordering today, so this is the `anchorsCommentBelow` hole reachable through a second pin reason rather than a regression. Say the word and I will file it.

## Verification

```
$ npm run compile

> tsc -p ./

$ npm test

> jest --verbose

Test Suites: 25 passed, 25 total
Tests:       508 passed, 508 total
Snapshots:   0 total
Time:        11.828 s, estimated 12 s
Ran all test suites.

$ npm run format:check

> prettier --check "**/*.ts"

Checking formatting...
```

**Regression proof.** With `src/imports/ImportScanner.ts` and `src/imports/ImportDocumentEditor.ts` stashed and the tests left in place, exactly the five new `ImportDocumentEditor` tests fail and the other 115 pass:

```
  ● ImportDocumentEditor.buildOrganizedContent › leaves a using: pair opened after a semicolon alone
  ● ImportDocumentEditor.buildOrganizedContent › organizes the other imports around a using: pair opened after a semicolon
  ● ImportDocumentEditor.buildOrganizedContent › does not write a second copy of a path such a pair already provides
  ● ImportDocumentEditor.buildOrganizedContent › leaves a using: pair alone when a comment trails the opener
  ● ImportDocumentEditor.buildOrganizedContent › leaves the line alone when such a using: opens no indented path
Tests:       5 failed, 115 passed, 120 total
```

## Review gate

Two rounds, fresh context each time.

- **Round 1: FAIL** - 1 Critical (a trailing comment on the opener defeated the fix entirely and restored the reported corruption), 4 Important, 3 Suggestion. The Critical, and two of the Importants that shared its root cause, are fixed here with tests; the changelog fragment it flagged as missing was added.
- **Round 2: PASS_WITH_SUGGESTIONS** - all round-1 findings verified closed, no new defect. Six suggestions; five applied. The declined one is collapsing the two `using:` branches into one, which is a refactor of code this ticket did not ask about.

The gate also corrected my reasoning on one point: I had justified testing the whole-line branch against the raw line by a case that `isModuleImport` already rejects at the gate. That comment now states the real reason.

### DIFF
--- a/changelog.d/219.fixed.md
+++ b/changelog.d/219.fixed.md
@@ -1,0 +1,1 @@
+**Import organizing**: a `using:` block opened after a semicolon keeps both its lines, instead of losing the opener and leaving its path stranded in the file body.
--- a/changelog.d/219.fixed.md
+++ b/changelog.d/219.fixed.md
@@ -1,1 +1,1 @@
-**Import organizing**: a `using:` block opened after a semicolon keeps both its lines, instead of losing the opener and leaving its path stranded in the file body.
+**Organize imports**: a `using:` pair opened after a `;` keeps both of its lines, instead of losing the opener and leaving its path stranded in the file body.
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -147,8 +147,10 @@ export class ImportDocumentEditor {
             // is what puts the marker somewhere it was not written, so it is
             // refused here rather than only by every caller passing a filtered
             // list. Every caller does today; one that forgot would resurrect a
-            // defect this branch has already shipped once.
-            if (!imp.trailingComment || imp.anchorsCommentBelow) {
+            // defect this branch has already shipped once. A statement sharing
+            // its span is refused for the same reason: what trails it is the
+            // rest of the span, not an annotation to re-emit.
+            if (!imp.trailingComment || imp.anchorsCommentBelow || imp.rebuildLosesText) {
                 continue;
             }
             const statement = this.formatter.formatImportStatement(imp.path, preferDotSyntax);
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -579,16 +579,19 @@ export class ImportDocumentEditor {
     ): string | null {
         const eol = detectEol(text) ?? options.fallbackEol ?? "\n";
         const lines = text.split(LINE_SPLIT);
-        // An import anchored by a comment marker is neither hoisted nor
-        // removed from the body: its line stays exactly where it is, and the
-        // rest of the file is organized around it. It is still an import,
-        // though, so an additional path it already covers must not be written
-        // out a second time - that would duplicate it rather than move it.
+        // A pinned import - one anchored by a comment marker, or one sharing its
+        // span with another statement - is neither hoisted nor removed from the
+        // body: its line stays exactly where it is, and the rest of the file is
+        // organized around it. It is still an import, though, so an additional
+        // path it already covers must not be written out a second time - that
+        // would duplicate it rather than move it. Both reasons answer that the
+        // same way, so the set is keyed on the line staying put rather than on
+        // why it does; see ScannedImport.rebuildLosesText.
         const allImports = scanModuleImports(lines);
         const movableImports = rewritableImports(allImports);
-        const anchoredPaths = new Set(allImports.filter((imp) => imp.anchorsCommentBelow).map((imp) => imp.path));
+        const pinnedPaths = new Set(allImports.filter((imp) => imp.anchorsCommentBelow || imp.rebuildLosesText).map((imp) => imp.path));
 
-        const extraPaths = additionalPaths.map((p) => p.trim()).filter((p) => p.length > 0 && !anchoredPaths.has(p));
+        const extraPaths = additionalPaths.map((p) => p.trim()).filter((p) => p.length > 0 && !pinnedPaths.has(p));
 
         // Module imports have file scope, so the names an anchored import brings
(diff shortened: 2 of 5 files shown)


vukefn:
The pre-existing ordering hole noted in the PR body is now tracked as #232. Both shapes verified: the anchored one reproduces independently of this change, the pinned one through the pin reason added here.

Issue #139, opened 2026-08-04
CRLF files get mixed line endings and Optimize Imports is never idempotent on them


vukefn:
Found in a codebase audit after v0.8.0. Verified by executing the writers against the real modules.

`buildOrganizedContent` (`ImportDocumentEditor.ts:295-327`) splits on `"\n"`, so body lines keep their trailing `\r` while the regenerated import lines are LF-only, and the result is rejoined with `"\n"`.

```
in : "using { /B }\r\nusing { /A }\r\ncode()\r\n"
out: "using { /A }\nusing { /B }\n\ncode()\r\n"
```

Worse, on a file that is **already organized**:

```
in : "using { /A }\r\n\r\ncode()\r\n"
out: "using { /A }\n\ncode()\r\n"      identical? false
```

so the `organized === text` short-circuit at `:349` never fires. `organizeImports` therefore does a full-document replace **plus `document.save()`** (`CommandsHandler.ts:133-135`) on every single invocation. On Windows, where `files.eol` defaults to `auto` and resolves to CRLF, that means a whole-file diff and a mixed-EOL buffer every time the user runs Optimize Imports.

`addImportsToDocument` has the same defect:

```
out: "using { /Fortnite.com/Devices }\nusing { /Verse.org/Simulation }\n\r\ncode()\r\n"
```

The scanner has a CRLF regression test (`ImportScanner.test.ts:52`); the writers have none.

Fix: detect the document's dominant EOL (or read `TextDocument.eol`) and use it when joining, rather than hardcoding `\n`.

PR #234, opened 2026-08-08
refactor(imports): derive the conversion range from the real line length

### LINKED ISSUE #150: ImportPathConverter splits on \n, so applyConversion's replace range overshoots by one on CRLF files
Follow-up from #139, which fixed the same class of bug in `ImportDocumentEditor`. `ImportPathConverter` is now the only writer path that still splits the document on `"\n"`.

Three sites split without handling the carriage return: `ImportPathConverter.ts:419`, `:512` and `:533`. The first two only ever use `line.trim()`, so the stray `\r` is discarded and nothing downstream is affected.

The third feeds a write. `applyConversion` builds its replace range from the split line's length:

```ts
const lines = text.split("\n");                       // :533
...
const range = new vscode.Range(new vscode.Position(lineIndex, 0), new vscode.Position(lineIndex, lines[lineIndex].length));   // :554
```

On a CRLF file `lines[lineIndex]` keeps its trailing `\r`, so the end column is one past the true end of the line.

**This is not currently a defect.** VS Code clamps an out-of-range position when it validates the edit, the replacement text contains no newline, and the line terminator sits outside the line's character range either way, so the conversion lands correctly and the CRLF survives. Verified by reading the path rather than by reproducing a failure: there is no input in this path that produces wrong output today.

What it costs is that the range is only right by accident. Correctness depends on host clamping rather than on the range being what the code says it is, and the file no longer matches the convention #139 established next door, where `ImportDocumentEditor.ts` splits on a shared `LINE_SPLIT` regex and joins with a detected line ending.

Proposed:

- Split on `/\r?\n/` at all three sites, so the range is derived from the real line length.
- Reuse the `LINE_SPLIT` constant rather than repeating the literal, exporting it from `ImportDocumentEditor.ts` or moving it somewhere both files can reach if that is the cleaner seam.

Scope:

- In: the three splits in `ImportPathConverter.ts` and the replace range they feed.
- Out: which line the conversion targets, which is #141. Also out: EOL detection, since this path writes no newline and needs none.

Acceptance:

- `applyConversion` on a CRLF document produces a replace range whose end column equals the line length without the carriage return.
- Converting an import in a CRLF file leaves every other line byte-identical, asserted by a regression test.

Collision, not a dependency: #141 rewrites how `applyConversion` (`:531-545`) finds its target line. Both changes edit the same handful of lines, so they should not run in separate worktrees at the same time, but neither blocks the other and either can land first.

vukefn:
Closes #150

`ImportPathConverter` split the document on a bare `"\n"` at three sites, so on a CRLF file every line kept its carriage return. The third of those feeds a write: `applyConversion` derives its replace range from the split line's length, which put the end column one past the true end of the line.

This is not a defect today, and that still holds against code added since the issue was filed — VS Code clamps the out-of-range column, and `ImportFormatter.extractTrailingComment` ends in `.trim()`, so the stray `\r` never reaches the rebuilt line either. What it cost is that the range was correct only because the host salvaged it, and the file contradicted the convention #139 established next door.

## Changes

- `LINE_SPLIT` moves from `ImportDocumentEditor.ts`, where it was unexported, to `ImportScanner.ts`, where it is exported. That is the line-level module both writers already import from, and every function it exports takes `lines: string[]` — the constant that produces them belongs beside them. Exporting from `ImportDocumentEditor` would instead have made the converter depend on a sibling writer. The regex carries no `g` flag, so it holds no `lastIndex` state and is safe to share.
- The converter's three sites split on it.
- The `findConversionLine` doc comment claimed the trimmed comparison "absorbs the trailing `\r` a CRLF document leaves on every element of a `"\n"` split" — no longer true, so it now says what trimming is still for.

## Tests

Two regression tests, both proven to fail against the unfixed code by stashing the source and running them. The vscode mock's `Range`/`Position` do not clamp, which is what makes the defect observable where real VS Code hides it:

- the replace range ends at column 23 rather than 24 on a CRLF line — the acceptance criterion's "end column equals the line length without the carriage return";
- applying the recorded operation to the CRLF text leaves every other line byte-identical. Unfixed, the overshooting range eats the `\r` and leaves a lone `\n`.

`fakeDocument` gained an optional line-ending argument to build them, and the existing `"does not carry the carriage return..."` test now constructs its CRLF document that way instead of hand-writing `\r` into each line — its comment described the `"\n"` split this change removes.

No changelog fragment: there is no user-visible effect.

## Out of scope

Which line the conversion targets (#141, already landed) and EOL detection, both excluded by the ticket.

The review gate flagged that `ImportCodeLensProvider.ts:164` splits on `"\n"` and builds a range from `lines[i].length` at `:178` — the same shape, in a lens range rather than a write. Confirmed and left alone as outside this ticket; it deserves its own issue.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
> jest --verbose

Tests:       538 passed, 538 total
Snapshots:   0 total
Time:        9.323 s
Ran all test suites.

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

Review gate: PASS_WITH_SUGGESTIONS, 0 Critical, 0 Important, 2 Suggestions. The first is addressed in `58d2ff4` — the new `LINE_SPLIT` doc asserted a module-wide precondition the code lens provider violates, so it now claims only what it can vouch for. The second is informational: the pre-existing trailing-comment test is protected by `.trim()` rather than by the split, so it is not coverage of this change.

### DIFF
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -1,7 +1,7 @@
 import * as vscode from "vscode";
 import { logger } from "../utils";
 import { ImportFormatter } from "./ImportFormatter";
-import { allUsingPaths, classifyLines, LineClassification, rewritableImports, scanModuleImports, ScannedImport } from "./ImportScanner";
+import { allUsingPaths, classifyLines, LINE_SPLIT, LineClassification, rewritableImports, scanModuleImports, ScannedImport } from "./ImportScanner";
 
 /** Represents a contiguous block of import statements in the document. */
 interface ImportBlock {
@@ -12,9 +12,6 @@ interface ImportBlock {
 
 /** The line ending a document is written with. */
 export type LineEnding = "\r\n" | "\n";
-
-/** Splits text into lines without keeping the carriage return of a CRLF pair. */
-const LINE_SPLIT = /\r?\n/;
 
 /**
  * Detects the dominant line ending of a text, or null when it has no line
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -4,7 +4,7 @@ import { logger } from "../utils";
 import { ProjectPathHandler } from "../project";
 import { ProjectPathCache } from "../services";
 import { ImportFormatter } from "./ImportFormatter";
-import { scanConvertibleImports } from "./ImportScanner";
+import { LINE_SPLIT, scanConvertibleImports } from "./ImportScanner";
 
 interface ImportConversionResult {
     originalImport: string;
@@ -450,7 +450,7 @@ export class ImportPathConverter {
     async convertAllImportsFromFullPath(document: vscode.TextDocument): Promise<ImportConversionResult[]> {
         const results: ImportConversionResult[] = [];
         const text = document.getText();
-        const lines = text.split("\n");
+        const lines = text.split(LINE_SPLIT);
 
         for (const { statement, line } of scanConvertibleImports(lines)) {
             if (this.isFullPathImport(statement) && !this.isBuiltinModule(statement)) {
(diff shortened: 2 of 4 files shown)


vukefn:
The same overshoot shape in `ImportCodeLensProvider` is now filed as #238.

Issue #144, opened 2026-08-04
Saving a Verse file can immediately re-dirty it


vukefn:
Found in a codebase audit after v0.8.0.

`extension.ts:151-161` — the `onDidSaveTextDocument` handler awaits `importHandler.ensureEmptyLinesAfterImports(document)`, which applies a `WorkspaceEdit` (`ImportDocumentEditor.ts:430-449`) and never re-saves.

Scenario: a file whose import block has 0 or 2 blank lines after it while `behavior.emptyLinesAfterImports` is 1. The user presses Ctrl+S. The file saves, then the extension edits it, and the tab immediately shows the unsaved dot. A second Ctrl+S converges, since the spacing now matches and `:425-428` short-circuits.

Two smaller things in the same handler:

- It is `async` with no try/catch, so anything thrown before `ensureEmptyLinesAfterImports`'s inner try (`:383-428`) becomes an unhandled rejection — `onDidSaveTextDocument` does not await listener promises.
- The `emptyLinesAfterImports >= 0` guard at `:156` is always true, since `package.json:161` sets `"minimum": 0`.

PR #240, opened 2026-08-09
feat(config): let emptyLinesAfterImports preserve existing spacing

### LINKED ISSUE #151: Add a way to opt out of the empty-lines-after-imports normalization
Reported on Twitter by @Juno_006: "would be nice if this wasn't forced". The screenshot shows `verseAutoImports.behavior.emptyLinesAfterImports` set to 0.

## Current behavior

- `extension.ts:151-161` runs `ensureEmptyLinesAfterImports` on every save of every `.verse` document, whether or not the extension did any import work in that file.
- `package.json:158-165` declares `"minimum": 0`, so no value expresses "leave the existing spacing alone". At 0 the pass deletes blank lines the user wrote between the import block and the first line of code.
- The `emptyLinesAfterImports >= 0` guard at `extension.ts:156` is therefore dead code. Also noted in #144.
- `ImportDocumentEditor.ensureEmptyLinesAfterImports` (`:377-449`) has no guard at all, so the add-import path (`:267`) and the organize path (`:363`) normalize spacing unconditionally.

## Proposal

`-1` means "leave existing spacing untouched".

- `package.json`: `"minimum": -1`, description updated to document the sentinel.
- `ImportDocumentEditor.ensureEmptyLinesAfterImports`: return early when the configured value is negative, so the add and organize paths honor it too.
- `extension.ts:156`: the existing guard becomes live rather than dead.

Default stays 1, so existing installs are unaffected.

## Notes

- Overlaps #144, which fixes the re-dirty and the unhandled rejection in the same save handler. Whichever lands second must not reintroduce the dead guard.
- Alternatives considered and rejected: a separate `normalizeEmptyLinesOnSave` boolean (covers only the save trigger, leaves add and organize still forcing the count), and a string enum with a `preserve` member (more discoverable in the settings UI, but changes the type of a shipped setting and needs a migration).

vukefn:
Closes #151

`behavior.emptyLinesAfterImports` declared `"minimum": 0`, so no value expressed "leave the spacing after my import block alone" — and `0`, the value the reporter reached for, deletes the blank lines the author wrote. This declares `-1` as the minimum and documents the sentinel.

## The ticket was half-landed already

#151 was filed before #144 merged, and #144 already added the negative-value early return that the ticket's second and third bullets ask for. So the runtime half exists on `main` today: a hand-edited `-1` in `settings.json` already preserves spacing on the save, add and organize paths, since all three reach `computeEmptyLinesAfterImportsEdits`. What was missing is that the settings UI refuses to store a value below the declared minimum, so the opt-out was unreachable and undocumented.

This PR is therefore only the schema, the comment that described the guard and is now wrong, and tests. It deliberately does not touch `src/extension.ts`, and does not change the guard's logic.

Two line references in the ticket body are stale and were not acted on: the save handler is now an `onWillSaveTextDocument` participant at `src/extension.ts:223-234`, and the dead `>= 0` guard the ticket asks to revive no longer exists in `extension.ts` at all.

## Changes

- `package.json` — `"minimum": -1`, description documents the sentinel. `default: 1` and `maximum: 5` are unchanged, so existing installs see no behaviour change.
- `src/imports/ImportDocumentEditor.ts` — comment and debug message only. The old comment explained the guard as defence against a settings file edited past the declared minimum; that reading is wrong once `-1` is supported.
- `src/imports/__tests__/ImportDocumentEditor.test.ts` — a test that existing blank lines survive at `-1` (the case the ticket is actually about; the existing negative test only covered the zero-blank-lines input), and one asserting the manifest admits `-1`, following the repo's existing idiom of reading `contributes.configuration.properties`.
- `changelog.d/151.added.md` — new fragment.
- `changelog.d/144.fixed.md` — dropped one sentence. It framed a negative value as unsupported input that gets ignored, which contradicts this PR's "set it to -1 to keep your spacing" in the same unreleased section. Both fragments assemble into one release body. Revert that hunk if you would rather keep it.

## On the regression tests

Being straight about what these demonstrate: only the manifest test fails without this diff. Taking the fix away proves it:

```
git stash push -- package.json
npx jest src/imports/__tests__/ImportDocumentEditor.test.ts -t "package.json admits"

    Expected: <= -1
    Received:    0

    > 1495 |         expect(setting.minimum).toBeLessThanOrEqual(-1);

Test Suites: 1 failed, 1 total
Tests:       1 failed, 138 skipped, 139 total
```

The two behavioural tests pass against unfixed code, because #144 already shipped that behaviour. They pin it rather than prove a fix.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Tests:       561 passed, 561 total
Snapshots:   0 total
Time:        7.801 s, estimated 9 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

Review gate: PASS_WITH_SUGGESTIONS, 0 critical, 1 important, 4 suggestions. The comment trim, the test name, and the `144.fixed.md` sentence are the suggestions applied.

## Follow-up, not in this PR

The wiki still documents this setting's range as `0-5` (`Configuration.md:39`, which `README.md:70` points at as the settings reference), so the opt-out stays undiscoverable for exactly the reported use case until that page is updated. It lives in the separate `.wiki` repo, so it cannot land here.

Left alone deliberately: `CHANGELOG.md:136` also says `0-5 lines`, but that is a released historical entry.

### DIFF
--- a/changelog.d/144.fixed.md
+++ b/changelog.d/144.fixed.md
@@ -1,1 +1,1 @@
-**Saving no longer re-dirties the file**: the import spacing pass now runs as part of the save rather than editing the file once the save has finished, so the tab no longer shows unsaved changes the moment you press Ctrl+S. A hand-edited negative `behavior.emptyLinesAfterImports` is ignored instead of removing a line of code.
+**Saving no longer re-dirties the file**: the import spacing pass now runs as part of the save rather than editing the file once the save has finished, so the tab no longer shows unsaved changes the moment you press Ctrl+S.
--- a/changelog.d/151.added.md
+++ b/changelog.d/151.added.md
@@ -1,0 +1,1 @@
+**Empty lines after imports**: setting `behavior.emptyLinesAfterImports` to -1 now leaves the spacing after your import block exactly as you wrote it, instead of normalizing it on every save, add and organize.
--- a/package.json
+++ b/package.json
@@ -212,9 +212,9 @@
         "verseAutoImports.behavior.emptyLinesAfterImports": {
           "type": "number",
           "default": 1,
-          "minimum": 0,
+          "minimum": -1,
           "maximum": 5,
-          "description": "Number of empty lines to add after the imports section at the top of the file",
+          "description": "Number of empty lines to keep after the imports section at the top of the file. Set to -1 to leave the existing spacing untouched.",
           "order": 14
         },
         "verseAutoImports.behavior.ambiguousImports": {
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -1047,11 +1047,11 @@ export class ImportDocumentEditor {
         const config = vscode.workspace.getConfiguration("verseAutoImports");
         const emptyLinesAfterImports = config.get<number>("behavior.emptyLinesAfterImports", 1);
 
-        // package.json declares a minimum of 0, but a hand-edited settings.json
-        // does not have to honour it, and a negative value reaches the delete
-        // branch below, where it would take real code rather than blank lines.
+        // A negative value means "leave the file's own spacing alone";
+        // package.json declares -1 for it. Without this return the negative
+        // difference below deletes real code.
         if (emptyLinesAfterImports < 0) {
-            logger.debug("ImportDocumentEditor", `Ignoring a negative emptyLinesAfterImports (${emptyLinesAfterImports})`);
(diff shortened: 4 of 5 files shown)


vukefn:
Follow-ups filed from this change, neither in scope here: #241 (wiki still documents the range as 0-5, so the -1 opt-out stays undiscoverable) and #242 (a fractional value makes the save participant apply a no-op edit on every save; predates this PR).

PR #243, opened 2026-08-09
fix(imports): pin a line that writes a definition after its import

### LINKED ISSUE #235: Organizing deletes a non-using statement written after an import on the same line
A `;` separates definitions exactly as a newline does, so an import can share its line with a definition that is not an import. Organizing deletes it:

```
in : using { /X }; MyVal := 5

     code()

out: using { /X }

     code()
```

`MyVal` is gone. As in #223, the output is a well-formed import block, so nothing signals the loss.

## Where

`scanModuleImports` (`src/imports/ImportScanner.ts`) records the line through `extractPathFromImport`, which reads the statement at the head of what it is given and reports nothing about the remainder. The entry spans the one line and is rewritable, so `hoistedLines` takes the line and the writer re-emits it from `/X` alone.

The two fixes already landed for this family both key on finding a *second `using`* on the line:

- #219 / #231 pin a line whose tail is a `using:` opening an indented pair.
- #223 pins a line that writes more than one complete `using` statement, counted by `usingPathsOnLine`.

A `;` followed by anything that is not a `using` matches neither, so the line still reaches the single-statement branch and is rebuilt from its one path. `ScannedImport.rebuildLosesText` is `false` for it, and that field's doc names this as one of the two shapes the flag still misses.

## Suggested acceptance

- Organizing `["using { /X }; MyVal := 5", "", "code()"]` leaves the line as written.
- A line whose `;` is inside a comment, a string, or a path is unaffected, and an ordinary single-statement line stays rewritable and organizable.

## The hard part

Detecting the shape needs a `;` that really is a separator, at depth 0 outside comments and string literals - `usingPathsOnLine` sidesteps that today by searching for `using` rather than for the separator. `classifyLines` already strips comments and hands back a `code` string, so the remaining question is string literals. Worth settling before implementing, as in #223.

Found during the review of #223, and left out of that PR because the shape is neither the one the ticket named nor reachable through the branch it added.

vukefn:
Closes #235

A `;` separates definitions exactly as a newline does, so an import can share its line with a definition that is not an import. Organizing deleted it:

```
in : using { /X }; MyVal := 5      out: using { /X }
```

The output was a well-formed import block, so nothing signalled the loss.

## The approach: no `;` scanner

The ticket flags the hard part — detecting a `;` that really is a separator means finding one at depth 0 outside comments and string literals, and nothing in `src/` tracks string state or brace depth.

Asking a different question removes the problem: **did the statement account for the whole line?** The braced pattern has a closing delimiter, so live text after it is text a rebuild would lose, whatever the separator is. The `;` itself is never searched for, which is what makes the three acceptance shapes fall out rather than needing cases:

- a `;` in a comment — `classifyLines` already hands back a comment-free `code` string, and the check is asked of that, never the raw line;
- a `;` inside the braces — captured as the path, nothing left over, still rewritable;
- a `;` in a string literal — it can only ever sit inside leftover text, never be what decides there is any.

## Changes

- `ImportFormatter`: one private `matchImport` now owns the dotted-then-braced patterns and additionally reports where the statement ends; `extractPathFromImport` reads its path (behaviour unchanged), and the new `textAfterImport` reads the rest.
- `ImportScanner`: the single-statement branch sets `rebuildLosesText` from that leftover. Every writer already reads `rewritableImports`, so pinning is the whole fix.
- The `ScannedImport.rebuildLosesText` doc named this shape as a known miss; rewritten for what detection now covers and what it still does not.

## Deliberately out of scope

The dotted style, `using. /X; MyVal := 5`, is unchanged and still rewritable: its capture is greedy to end of line, so the statement swallows the tail into its path and nothing is left over to detect. It corrupts rather than deletes, it is the separately documented `usingPathsOnLine` weakness rather than the shape this ticket names, and the field doc now records it accurately. Worth its own issue.

A line ending in a bare `;` is now pinned, on text that means nothing. That is the safe direction — the cost is one line left as written — and trimming separators away before the test means being right about which ones mean nothing.

## Tests

Regression tests in `src/imports/__tests__/`, all without the VS Code runtime: the ticket's shape pinned and not rewritable at the scanner, left as written at the organize level with the other imports organizing around it, `textAfterImport` covered directly, plus negative controls for a `;` in a trailing comment and a `;` inside the braces.

Proven to have power — with `ImportScanner.ts` and `ImportFormatter.ts` stashed, 6 of the new cases fail and the `textAfterImport` suite does not compile. The two negative controls pass either way, which is what makes them controls.

Reviewed by a fresh reviewer: PASS_WITH_SUGGESTIONS, 0 Critical, 0 Important, 5 Suggestions. Four were applied (the doc inaccuracy above, comment trimming, a parameter renamed to carry its precondition, one over-broad claim tightened); the fifth records the bare-`;` narrowing as accepted.

## Verification

```
> tsc -p ./

=== TEST ===

Tests:       571 passed, 571 total
Snapshots:   0 total
Time:        9.146 s
Ran all test suites.
=== FORMAT ===

Checking formatting...
```

### DIFF
--- a/changelog.d/235.fixed.md
+++ b/changelog.d/235.fixed.md
@@ -1,0 +1,1 @@
+**Organize imports**: A line that writes a definition after an import, such as `using { /X }; MyVal := 5`, is now left as written instead of losing the definition.
--- a/src/imports/ImportFormatter.ts
+++ b/src/imports/ImportFormatter.ts
@@ -170,8 +170,8 @@ export class ImportFormatter {
      * because the comment is only stripped afterwards, from that capture.
      *
      * One home for the two patterns, so a caller asking where the statement
-     * ends cannot answer it from a looser second copy that disagrees with the
-     * path about which statement was read.
+     * ends cannot answer it from a looser second copy that, on the same input,
+     * disagrees about which statement was read.
      *
      * @returns The path and the offset just past the statement **in the
      *   trimmed input**, or null when there is no statement (including one
@@ -214,11 +214,13 @@ export class ImportFormatter {
      * the statement is the whole of it.
      *
      * A writer rebuilds an import line from its path, so whatever this reports
-     * is text that rebuild would delete. That makes it the question to ask
-     * about a line carrying a `;`: the separator itself is never searched for,
-     * so a `;` inside the braces or in a path is part of the statement and
-     * reported as nothing, and no string-literal tracking is needed to tell a
-     * separator from a character in a literal.
+     * is text that rebuild would lose. Asking it this way rather than searching
+     * the line for the `;` that separates two statements is what keeps a
+     * separator apart from a `;` that is only a character - one inside the
+     * braces, or in a path, is part of the statement and leaves nothing over,
+     * and one inside a string literal cannot arise without leftover text around
+     * it. So no depth or string tracking is needed, which nothing in this file
+     * does.
      *
      * Only the braced style can answer usefully. The dotted style has no
      * closing delimiter and its capture is greedy to end of line, so a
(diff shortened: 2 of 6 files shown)


vukefn:
The dotted shape this PR leaves out is now tracked as #244.

PR #246, opened 2026-08-09
fix(imports): end a dotted import where its path ends

### LINKED ISSUE #244: Organizing folds a statement written after a dotted import into its path
The dotted style has no closing delimiter, so a statement sharing its line swallows everything after it into its path. Organizing then rebuilds the line from that path:

```
in : using. /X; MyVal := 5

     code()

out: using { /X; MyVal := 5 }

     code()
```

`MyVal` is not deleted, as it was in #235 - it is folded into the import, and the file no longer compiles. Under dot-syntax preference the text survives, but the line is still free to move and to sort as though `/X; MyVal := 5` were a path.

## Where

`extractPathFromImport` (`src/imports/ImportFormatter.ts`) matches the dotted style with `^using\.\s*(.+)`, greedy to end of line. Both of `scanModuleImports`'s guards are blind to what that produces:

- `usingPathsOnLine` counts one `using` on the line, so the multi-statement branch (#223) does not fire.
- `textAfterImport` reports nothing left over, because the capture already took it, so the leftover branch (#235) does not fire either.

The entry lands on the single-statement branch with `rebuildLosesText: false` and is rewritable. `ScannedImport.rebuildLosesText`'s doc names this as the one shape both questions miss.

Where the tail is itself a `using`, the line *is* pinned - by the second `using`, not by anything that understood the swallow - but under the swallowed path, so the path it really imports goes unreported and a writer may add a second copy of it. `scanConvertibleImports` also still offers a conversion CodeLens on the rewritable case.

## Suggested acceptance

- Organizing `["using. /X; MyVal := 5", "", "code()"]` leaves the line as written.
- `using. /X; using { Economy.Shop }` reports `/X` as a path in its own right, so nothing writes a second copy of it. It is already pinned; that does not change.
- An ordinary dotted statement, `using. /Verse.org/Simulation`, with or without a trailing comment, stays rewritable and organizable, and its path is unchanged.

## The hard part

Giving the dotted style an end means deciding where a dotted path stops, which the pattern deliberately does not do today - a path has no closing delimiter and may hold `.`, `/`, `@` and `-`. A separator test at that boundary is the `;` question #235 avoided by asking for leftover text instead, and here there is no way to avoid it. Worth settling before implementing.

Found during the review of #235; see PR #243, which deliberately leaves this shape out.

vukefn:
Closes #244

The dotted style closes with nothing, so `matchImport` read its content to end of line and took a statement written after a `;` into the import path. Both of `scanModuleImports`' guards were blind to the result, because the greedy capture had already consumed the evidence each looks for: one `using` on the line, and nothing left over. The entry reached the single-statement branch as rewritable, and organizing rebuilt the line from that path into `using { /X; MyVal := 5 }` — the definition folded inside the braces, and the file stopped compiling.

The capture now ends where a dotted content ends, which is how Verse itself lexes a path.

## Where a dotted path stops

The ticket flagged this as the thing to settle first, so it was settled against the language's own corpus rather than by choosing a separator:

- `Tests/Literals/Path.versetest:3` carries the path production, and `:13-15` states outright that a character outside it ends the path and the remainder parses as a separate compound expression — `/a/b-c` is the path `/a/b`, then `-`, then `c`. Path lexing is maximal munch.
- `docs/01_expressions.md:1241` confirms `;` separates definitions in a scope, so a second statement on the line must be preceded by one.
- `Tests/QualifiedIdentifier.versetest:70` and `:300` show the `(path:)` qualifier used inside a real `using`, which is why `(`, `)` and `:` are in the class.
- A quoted segment suffix, `Economy.Shop'Loc'`, admits characters no identifier may hold, so it is consumed whole rather than stopped at.

What the Book does **not** settle is whether `using. /X; MyVal := 5` parses as two definitions or as one malformed `using` whose argument is a sequence — the dotted `using` is undocumented and appears nowhere in the corpus. That does not block the fix: leaving the line as written is correct under both readings, and under the second the line is already invalid Verse, since `using` takes exactly one argument (`Tests/Modules/Using.versetest:68`).

## Acceptance

| Input | Result |
| --- | --- |
| `using. /X; MyVal := 5` | path `/X`, leftover `; MyVal := 5`, line pinned; organize returns unchanged and no conversion lens is offered |
| `using. /X; using { Economy.Shop }` | `/X` reported in its own right alongside `Economy.Shop`, so nothing writes a second copy of it; still pinned |
| `using. /Verse.org/Simulation`, with or without a trailing comment | unchanged, still rewritable and organizable |

Regression tests are in `src/imports/__tests__/`, which run without the VS Code runtime. Each was proved to detect the defect: reverting `ImportFormatter.ts` to its pre-fix version fails 11 tests, restoring it returns the suite to green.

## Review gate

Two rounds, both against a fresh reviewer with its own context.

**Round 1 — FAIL, one Critical, accepted and reproduced before acting.** Ending the capture at the first character outside a bare path truncated two legal path forms, and truncation turned out not to be the conservative direction it looked. `using. Economy.Shop'Loc'` recorded the path as `Economy.Shop` — a different module that exists — so a writer asked for `Economy.Shop` read the file as already importing it and withheld the import the compiler asked for. Separately `using. (/A:)B.C` became invisible to `allUsingPaths`, whose contract names a missed `using` as the one error it must not make; that was a regression this branch introduced against `main`, so it was fixed in code rather than documented as a gap.

**Round 2 — PASS_WITH_SUGGESTIONS.** Both suggestions concerned doc comments that had stopped being true once the class widened, and both are fixed in `713c470`. The reviewer independently confirmed the class admits exactly `()-./0-9:@A-Za-z_` with no unintended ranges and no backtracking risk, and that seven adversarial shapes are byte-identical to `main`.

## Scope

Two sibling copies of the same greedy pattern are deliberately untouched, per the maintainer's call:

- `ImportFormatter.isModuleImport` — consequence is that `using. Features; MyVal := 5` is classified as not-an-import and skipped whole. Non-destructive, but the path goes uncounted, so a writer may add a second copy of it. Worth a follow-up issue; the reproduction is in hand.
- `ImportPathConverter.extractPathFromImport` — unreachable with a swallowing statement, since every statement it sees arrives through `scanConvertibleImports`, which filters out pinned entries.

The braced branch is unchanged, `using { /X;Y }` included. Braces delimit, so the scanner can decline to parse inside them; the dotted form has no delimiter, so it must.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
> jest --verbose

Tests:       590 passed, 590 total
Snapshots:   0 total
Time:        12.93 s
Ran all test suites.

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/244.fixed.md
+++ b/changelog.d/244.fixed.md
@@ -1,0 +1,1 @@
+**Import organizing**: a statement written after a dotted `using.` import on the same line is no longer folded into the import path, and the line is left exactly as written.
--- a/src/imports/ImportFormatter.ts
+++ b/src/imports/ImportFormatter.ts
@@ -160,6 +160,25 @@ export class ImportFormatter {
     }
 
     /**
+     * A dotted `using` statement, capturing its content.
+     *
+     * The content class is the union of the three things a dotted `using` can
+     * name: an absolute path, a dot-notation module reference such as
+     * `Economy.Shop`, and a bare folder-module identifier. Verse's path
+     * production allows two further forms this omits on purpose - a `'...'`
+     * segment suffix, which admits almost every printable character, and a
+     * `(path:)` qualifier - because admitting either means parsing where this
+     * only lexes, and truncating one costs the line nothing but its
+     * rewritability.
+     *
+     * A comment opener cannot appear in the class, so a trailing comment ends
+     * the capture rather than entering it. matchImport strips one from the
+     * capture regardless, so widening the class cannot quietly put a comment
+     * back into a path.
+     */
+    private static readonly DOTTED_STATEMENT = /^using\.\s*([A-Za-z0-9_./@-]*)/;
+
+    /**
      * The `using` statement written at the head of a string: its path, and how
      * much of the string it occupies.
      *
@@ -173,6 +192,13 @@ export class ImportFormatter {
      * ends cannot answer it from a looser second copy that, on the same input,
      * disagrees about which statement was read.
      *
+     * The dotted style closes with nothing, so its own content is what ends it:
+     * the capture stops at the first character a dotted content cannot hold,
+     * which is how Verse itself lexes a path - `/a/b-c` is the path `/a/b`, then
+     * `-`, then `c`. Reading to end of line instead is what let a statement
+     * written after a `;` be captured into the path and re-emitted inside the
+     * braces. See DOTTED_STATEMENT for what a dotted content may hold.
+     *
      * @returns The path and the offset just past the statement **in the
      *   trimmed input**, or null when there is no statement (including one
      *   whose content is nothing but a comment)
@@ -180,7 +206,7 @@ export class ImportFormatter {
     private static matchImport(importStatement: string): { path: string; end: number } | null {
         const trimmed = importStatement.trim();
 
-        const dotMatch = trimmed.match(/^using\.\s*(.+)/);
+        const dotMatch = trimmed.match(ImportFormatter.DOTTED_STATEMENT);
         if (dotMatch) {
             const path = ImportFormatter.stripTrailingComment(dotMatch[1]);
             return path ? { path, end: dotMatch[0].length } : null;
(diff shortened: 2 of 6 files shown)


vukefn:
Filed the scoped-out gap as #247: `isModuleImport` keeps its own greedy dotted pattern, so `using. Features; MyVal := 5` is classified as not-an-import and skipped whole. Non-destructive, but the path goes uncounted, which lets a writer add a second copy of it.

Issue #225, opened 2026-08-08
comments: scanning, formatting and extraction (pilot for the src/ pass)


vukefn:
## Parent

#224

## What to build

The comment standard applied to import scanning, formatting and suggestion extraction: `ImportScanner`, `ImportFormatter`, `ImportSuggestionExtractor`, and the `ImportHandler` facade with the module barrel.

This ticket is first on purpose. It holds the worst material in the repository and the best, so the merged diff becomes the reference the remaining four tickets calibrate against.

The worst: `ScannedImport.anchorsCommentBelow` carries roughly twenty lines of doc for one boolean, and `LineScan.code` carries sixteen that defend a design choice by citing two named Book of Verse test corpora. Both encode a real constraint. Keep the constraint, drop the defence and the citations, and let the issue hold the argument.

The best: the `ImportHandler` facade documents why one method has no applying counterpart, which is a fact no reader can recover from the code. It is the target, not material to rewrite.

Watch the two meanings of `using` while editing here. A comment that blurs module imports and local-scope using is worse than none, because that distinction has already caused real bugs.

The scanner's comments describing Verse block-comment nesting and the indented marker document genuine language rules the code implements but does not teach. Tighten them; do not remove them.

## Blocked by

- None, can start immediately.

## Acceptance criteria

- [ ] Every comment in the file set conforms to the standard: contracts, invariants, non-obvious rationale and load-bearing ordering are kept; restated signatures, restated field names, restated lines, restated branches, refactor history and commented-out code are gone.
- [ ] Incident-shaped comments are rewritten down to the constraint they encode. The story of the bug does not survive; the rule a future editor must not break does.
- [ ] No issue or pull request numbers remain in the comments. The single exception is a workaround for an upstream bug, where the link is the comment's expiry date.
- [ ] Every exported symbol in the file set has a doc comment opening with one present-tense sentence saying what it produces. `@param` appears only where it carries a fact the type cannot.
- [ ] Nothing but comments changed. No renames, no extracted functions, no reordering, no type changes.
- [ ] Where a comment exists only because a name is poor, the comment stays and a follow-up refactor ticket is filed. Note the ticket number in the pull request.
- [ ] `npm run compile` passes.
- [ ] `npm test` passes, with no test file added, removed or edited.
- [ ] `npm run format:check` passes.
- [ ] The strip-diff gate is empty, and its output is pasted in the pull request.

## How to verify nothing but comments changed

[...]

PR #251, opened 2026-08-09
docs(services): apply the comment standard to digest and project-path services

### LINKED ISSUE #227: comments: digest and project-path services
## Parent

#224

## What to build

The comment standard applied to all of `services`: the digest parsing and lookup path, and the project-path cache and scanner.

`ProjectPathScanner` carries five doc blocks across 382 lines and is the main gap here. The cache and the parsers own state and invalidation rules that the signatures do not express, so this ticket is more about adding real contracts than deleting noise.

Read the merged diff from the blocking ticket first and match its calibration.

## Blocked by

- The scanning, formatting and extraction pass, for calibration.

## Acceptance criteria

- [ ] Every comment in the file set conforms to the standard: contracts, invariants, non-obvious rationale and load-bearing ordering are kept; restated signatures, restated field names, restated lines, restated branches, refactor history and commented-out code are gone.
- [ ] Incident-shaped comments are rewritten down to the constraint they encode. The story of the bug does not survive; the rule a future editor must not break does.
- [ ] No issue or pull request numbers remain in the comments. The single exception is a workaround for an upstream bug, where the link is the comment's expiry date.
- [ ] Every exported symbol in the file set has a doc comment opening with one present-tense sentence saying what it produces. `@param` appears only where it carries a fact the type cannot.
- [ ] Nothing but comments changed. No renames, no extracted functions, no reordering, no type changes.
- [ ] Where a comment exists only because a name is poor, the comment stays and a follow-up refactor ticket is filed. Note the ticket number in the pull request.
- [ ] `npm run compile` passes.
- [ ] `npm test` passes, with no test file added, removed or edited.
- [ ] `npm run format:check` passes.
- [ ] The strip-diff gate is empty, and its output is pasted in the pull request.

## How to verify nothing but comments changed

Compile both revisions with comments stripped and compare. An empty diff is the pass condition; this was verified on `ImportScanner` before the spec was written, where 284 deleted comment lines plus an inserted doc block still produced byte-identical output.

```
tsc -p ./ --removeComments --sourceMap false --outDir <base-out>   # on main
tsc -p ./ --removeComments --sourceMap false --outDir <head-out>   # on the branch
diff -r <base-out> <head-out>
```

No changelog fragment: this is not a user-facing change.

vukefn:
Closes #227

Applies the comment standard to `src/services`: the digest parsing and lookup path, and the project-path cache and scanner. Calibrated against the merged diff from #225 (`f2da720`), which is this branch's base.

Seven of the eight files changed. `digestParsing.ts` is untouched: its regex and parser docs already conform, and churning them would have cost the diff its reviewability for nothing.

## What changed, by shape

**Restated signatures, lines and branches deleted.** `PrecompiledDigestLoader` carried nine docs of the `Get a digest entry by identifier` shape; `ProjectPathScanner` carried a `// Check for <X> declaration` label above each of seven pattern branches. Both sets are gone.

**Contracts added where the signature could not carry them.** `scanProject` returns null both when no project is found and when the scan fails, and a caller cannot tell those apart. `parseVerseFile` returns `[]` on failure rather than throwing. `extractDeclarations` emits a 1-based `sourceLine` and a `fullPath` prefixed by the enclosing module chain. `DigestParser` gained a class doc for the two-tier contract its three methods spread out, including that the fallback to runtime parsing is one-way for the life of the instance.

**Load-bearing order kept and sharpened.** The variable branch must stay last: the `:` its pattern needs is also the first character of the `:=` that opens a type declaration, so it matches those too, and only the guards below it keep them out. Verified rather than asserted.

**One ordering comment deleted as false.** `// Check for extension method first (before regular function)` claimed an ordering constraint that does not exist: both patterns are `^`-anchored, and the function pattern cannot match a receiver-prefixed line at all. Checked directly against both regexes before removing it.

**Verse semantics verified, not trusted.** The surviving module-visibility comment asserts that a declaration with no specifier is `internal`. Confirmed against the Book of Verse, `docs/12_access.md` — the `<internal>` specifier is the default access level when none is given.

## Second commit: six claims the review caught

The review gate found that six comments I had written did not hold against the code. Each is corrected in `453674e`, and each was re-verified against the source before the fix:

- The data-directory fallback described the packaged layout **backwards**. `.vscodeignore` excludes `src/**` but re-includes `!src/data/**`, and nothing copies the JSON into `out/` — there is no copy step in `package.json` and `tsc` does not emit JSON, so `out/data` is absent after a clean compile. `src/data` is the live location in both a checkout and the `.vsix`; the `out/data` branch is dormant.
- Loaded state is **at least one file**, not all-or-nothing: `successCount > 0` latches it, so `isLoaded()` can be true over a partial index.
- Lookup order does **not** survive into the quick-fix menu: `ImportCodeActionProvider.sortSuggestions` re-sorts alphabetically when `quickFix.sortAlphabetically` is on.
- A file the scanner cannot read commits as **empty** rather than keeping its previous declarations, because `parseVerseFile` swallows its own errors and returns `[]`. The `failedFiles` path is reachable only for a throw raised outside it.
- `sourceLine` is 1-based and therefore is **not** a `vscode.Position` line.
- The asset-class segment **ends** the module path rather than starting the member run, and `includePrivate` lifts the module visibility filter as well as the private check.

## Acceptance criteria

- [x] Comments conform: contracts, invariants, non-obvious rationale and load-bearing ordering kept; restated signatures, field names, lines, branches, refactor history and commented-out code gone.
- [x] Incident-shaped comments reduced to their constraint.
- [x] No issue or PR numbers in comments.
- [x] Exported symbols carry a doc opening with one present-tense sentence. See the note below on public methods.
- [x] Nothing but comments changed.
- [x] No follow-up refactor ticket needed: no surviving comment exists solely to compensate for a poor name.
- [x] `npm run compile`, `npm test`, `npm run format:check` pass.
- [x] Strip-diff gate empty.

**On "every exported symbol":** read as the exported class, function or const, not every public member. Six delegating or self-describing methods lost their docs outright (`getEntry`, `getModuleIdentifiers`, `isLoaded`, `getLoadError`, and the two `getStats`). This matches the calibration: `f2da720` deleted the same shape from `ImportHandler`'s five pass-through methods while keeping a rich doc on the class. Say the word if you read the criterion the other way and I will restore them as real docs.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
> jest --verbose

Tests:       595 passed, 595 total
Snapshots:   0 total
Time:        8.604 s
Ran all test suites.

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

No test file was added, removed or edited.

### Strip-diff gate

```
$ npx tsc -p ./ --removeComments --sourceMap false --outDir <head-out>     # on 453674e
$ git checkout f2da720 -- src/services
$ npx tsc -p ./ --removeComments --sourceMap false --outDir <base-out>     # on the merge base
$ git checkout HEAD -- src/services
$ diff -r <base-out> <head-out>
strip-diff exit=0
```

Empty, across all 63 emitted `.js` files. This is the assertion that covers "nothing but comments changed"; the green suite alone would not prove it.

No changelog fragment: not a user-facing change.

### DIFF
--- a/src/services/AssetsDigestParser.ts
+++ b/src/services/AssetsDigestParser.ts
@@ -29,8 +29,8 @@ const INSTANCE_DECL = /^(\w+)(?:<[^>]*>)*\s*:\s*\w+\s*=\s*external\b/;
  * Assets.digest.verse.
  *
  * These names are what tells a dotted suggestion where the module ends: in
- * "Did you mean X.Y.Z.ClassName", a segment naming an asset class means
- * everything after it addresses members rather than modules.
+ * "Did you mean X.Y.Z.ClassName", a segment naming an asset class ends the
+ * module path, and that segment onward addresses members rather than modules.
  */
 export class AssetsDigestParser {
     private classNames: Set<string> = new Set();
--- a/src/services/AssetsDigestParser.ts
+++ b/src/services/AssetsDigestParser.ts
@@ -25,9 +25,12 @@ const CLASS_OR_STRUCT_DECL = /^(\w+)(?:<[^>]*>)*\s*:=\s*(?:class|struct)\b/;
 const INSTANCE_DECL = /^(\w+)(?:<[^>]*>)*\s*:\s*\w+\s*=\s*external\b/;
 
 /**
- * Parses the project's Assets.digest.verse file to extract class names.
- * This is used to determine the correct module boundary when inferring imports
- * from "Did you mean X.Y.Z.ClassName" error messages.
+ * The set of asset type names UEFN generated for this project, read from its
+ * Assets.digest.verse.
+ *
+ * These names are what tells a dotted suggestion where the module ends: in
+ * "Did you mean X.Y.Z.ClassName", a segment naming an asset class means
+ * everything after it addresses members rather than modules.
  */
 export class AssetsDigestParser {
     private classNames: Set<string> = new Set();
@@ -52,8 +55,13 @@ export class AssetsDigestParser {
     ) {}
 
     /**
-     * Gets the path to the project's Assets.digest.verse file.
-     * Location: {LOCALAPPDATA}\UnrealEditorFortnite\Saved\VerseProject\{ProjectName}\{ProjectName}-Assets\Assets.digest.verse
+     * The project's Assets.digest.verse, or null when there is no project name
+     * or no file at either known location.
+     *
+     * The file lives outside the workspace, under
+     * `{LOCALAPPDATA}\UnrealEditorFortnite\Saved\VerseProject\{ProjectName}\`,
+     * in a `{ProjectName}-Assets` subfolder on current UEFN and directly in the
+     * project folder on older versions.
      */
     async getAssetsDigestPath(): Promise<string | null> {
         if (this.cachedDigestPath && fs.existsSync(this.cachedDigestPath)) {
@@ -66,10 +74,8 @@ export class AssetsDigestParser {
             return null;
         }
 
-        // Get LocalAppData path
         const localAppData = process.env.LOCALAPPDATA || path.join(os.homedir(), "AppData", "Local");
 
-        // Try primary path structure (newer UEFN versions)
         const primaryPath = path.join(localAppData, "UnrealEditorFortnite", "Saved", "VerseProject", projectName, `${projectName}-Assets`, "Assets.digest.verse");
 
         if (fs.existsSync(primaryPath)) {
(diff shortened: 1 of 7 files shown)


vukefn:
Two pre-existing issues surfaced while commenting this code, filed rather than fixed here since this PR changed comments only: #255 (the cache drops a file's declarations when it cannot be read) and #256 (the unreachable out/data fallback). The doc comment on `invalidateFiles` describes the current behaviour; #255 is what would let it describe a transaction again.

Issue #224, opened 2026-08-08
Comment-quality pass across src/: apply one written standard to 34 non-test files


vukefn:
## Problem Statement

The comments in `src/` have drifted in two opposite directions at once, and both cost the same thing: a reader cannot tell, at a glance, what a symbol is for.

In the heavily worked areas - import scanning, document editing, path conversion - comments have accumulated as a record of the bugs that produced them. They read as transcribed pull-request arguments: a single boolean field carries twenty lines of doc, a single string field carries sixteen more that defend a design choice by citing two named Book of Verse test corpora. The durable half of that writing is a real constraint a future editor must not break. It is buried in argument, and the argument is addressed to an objection nobody reading the file has raised. Across `src/` roughly fifty comments carry this narrative shape, recognisable by "used to", "which is why", "the reason", and "previously".

Elsewhere the opposite holds. `StatusBarHandler` carries one doc block across 526 lines. `ImportCodeActionProvider` has none across 85. `DiagnosticsHandler` has four across 266, and `ProjectPathScanner` five across 382. These are not trivial files, and several of them own state whose rules are invisible in the signatures.

Between the two extremes sits a third category that is pure noise: comments that restate the line below them. A doc comment saying "Extracts existing import statements from a document" above a method named `extractExistingImports` taking a document. A field comment saying "The type of declaration" above a field named `type`. A branch comment saying "Normal state - just show text" inside an `else` that sets text. Each one trains the reader to skim comments, which is what makes the load-bearing ones dangerous, because the reader skims those too.

The maintainer works this codebase largely through coding agents. An agent reading a twenty-line defence of a boolean spends its context on the argument rather than the constraint, and an agent reading a file with no doc comments infers a contract from the implementation and then preserves whatever it inferred.

## Solution

One pass across the non-test source, applying a single written standard, so that every comment in `src/` says something the code cannot say on its own.

The standard already exists and is deliberately not part of this deliverable. It lives outside the repository at `~/.claude/rules/comments.md`, uncommitted, because the maintainer wants it to govern how agents write comments without becoming a public project convention that outside contributors are held to. It loads into every agent session automatically. Its spine:

- Keep contracts, invariants, non-obvious rationale, and load-bearing ordering.
- Delete restated signatures, restated field names, restated lines, restated branches, refactor history, and commented-out code.
[...]

PR #253, opened 2026-08-09
docs: apply the comment standard to activation, project, types and utils

### LINKED ISSUE #229: comments: activation, project, types, utils and scripts
## Parent

#224

## What to build

The comment standard applied to the remainder: the extension entry point, `ProjectPathHandler`, the type modules, the logger and environment utilities, and the digest parsing script.

The entry point leads this ticket and is the densest single example of both failure modes in the repository: 68 line comments across 301 lines, mixing pure noise such as a comment announcing that the logger is initialised with genuinely load-bearing rationale about why a disabled cache is left undefined rather than constructed, and why a stored payload is dropped at activation.

The project cache types carry twenty-one doc blocks across eighty-three lines. Some are among the best comments in the repository, such as what a `fullPath` actually spans and what bumping the cache version invalidates. Others say "The type of declaration" above a field named `type`. Density is not the measure; judge each block on whether it adds a fact.

Read the merged diff from the blocking ticket first and match its calibration.

## Blocked by

- The scanning, formatting and extraction pass, for calibration.

## Acceptance criteria

- [ ] Every comment in the file set conforms to the standard: contracts, invariants, non-obvious rationale and load-bearing ordering are kept; restated signatures, restated field names, restated lines, restated branches, refactor history and commented-out code are gone.
- [ ] Incident-shaped comments are rewritten down to the constraint they encode. The story of the bug does not survive; the rule a future editor must not break does.
- [ ] No issue or pull request numbers remain in the comments. The single exception is a workaround for an upstream bug, where the link is the comment's expiry date.
- [ ] Every exported symbol in the file set has a doc comment opening with one present-tense sentence saying what it produces. `@param` appears only where it carries a fact the type cannot.
- [ ] Nothing but comments changed. No renames, no extracted functions, no reordering, no type changes.
- [ ] Where a comment exists only because a name is poor, the comment stays and a follow-up refactor ticket is filed. Note the ticket number in the pull request.
- [ ] `npm run compile` passes.
- [ ] `npm test` passes, with no test file added, removed or edited.
- [ ] `npm run format:check` passes.
- [ ] The strip-diff gate is empty, and its output is pasted in the pull request.

## How to verify nothing but comments changed

Compile both revisions with comments stripped and compare. An empty diff is the pass condition; this was verified on `ImportScanner` before the spec was written, where 284 deleted comment lines plus an inserted doc block still produced byte-identical output.

```
tsc -p ./ --removeComments --sourceMap false --outDir <base-out>   # on main
tsc -p ./ --removeComments --sourceMap false --outDir <head-out>   # on the branch
diff -r <base-out> <head-out>
```

No changelog fragment: this is not a user-facing change.

vukefn:
Closes #229

Fifth and last slice of the comment-quality sweep tracked by #224, following the pilot in #249 and matching its calibration: a method whose name and signature answer the question gets no doc, and a paragraph has to be paying for a reader who would otherwise get it wrong.

Eleven files: `src/extension.ts`, `src/project/ProjectPathHandler.ts`, `src/project/index.ts`, `src/types/{configuration,index,moduleInfo,projectCache}.ts`, `src/utils/{environment,index,logger}.ts`, `src/scripts/parseDigestFiles.ts`. Seven changed. The four untouched are a deliberate no-op: the three barrels are re-export lines with nothing to say, and `utils/environment.ts` was already written to this standard.

## What changed

Deleted the restatement layer: the section headers in `activate` (`// Initialize the logger`, `// Create core services`, `// Set up file watchers`), the signature restatements on every `ProjectPathHandler` method and every `Logger` level method, and the field restatements in `projectCache.ts` (`/** The type of declaration */` on `type`).

Kept and tightened what carries a constraint: the undefined project path cache, the dropped stored payload, the save participant that must run during the save, the diagnostics scope snapshot, `fullPath`, and `PROJECT_CACHE_VERSION`.

Added contracts the types cannot carry - and the useful half of this diff is that writing them meant checking each against a call site:

- `sourceLine` is **one-based** (`i + 1` in `ProjectPathScanner`).
- `generatedAt` is milliseconds since the epoch, stamped by a full scan and by invalidation. A file deletion prunes `nodes` without restamping it, so it is the age of the last scan rather than a last-modified time.
- `ProjectScanOptions.includePatterns` - **only the first is read**; the scanner passes one include pattern to `findFiles` and drops the rest.
- `includePrivate` drops private declarations; modules alone are additionally held to public-or-internal.
- `ImportSuggestion.modulePath` is extracted back out of `importStatement` and is undefined when extraction fails, so callers must not assume it is present.
- `findAndParseProjectFile` caches only successes; a call that returns null repeats the whole search including the five-level parent walk.
- `Logger` writes INFO and above to the user channel and every level to the debug channel and the export buffer, and the two channels timestamp in different clocks (local wall-clock vs UTC to the millisecond).

## Two findings worth a maintainer's eye

**`src/types/configuration.ts` has no reader.** Every settings read in `src/` goes through `getConfiguration("verseAutoImports").get<T>(key, default)` with an inline type argument, so `AutoImportScope`, `GeneralConfig`, `BehaviorConfig`, `QuickFixConfig`, `PathConversionConfig`, `ExperimentalConfig` and `VerseAutoImportsConfig` describe the settings contract without participating in it. They have already drifted: the `cache` section has no interface at all. Documented as such rather than left looking live.

**Four types in `src/types/moduleInfo.ts` have no consumer anywhere in `src/`**: `ModuleInfo`, `MultiOptionStrategy`, `UnknownIdentifierResolution` and `QuickFixOrdering`. `ModuleInfo`'s field names (`intermediatePath`, `outerModule`, `internalModule`) do not carry enough to reconstruct what each segment spans, so rather than invent a contract the doc says that and warns against inferring one from the names.

Deleting either set is a type change and out of scope here. Happy to file a follow-up cleanup ticket if you want it.

No comment in the file set exists only to compensate for a poor name, so no refactor follow-up is required by that criterion.

## Deviation from one acceptance criterion, stated rather than skipped

"Every exported symbol in the file set has a doc comment" - the exported symbols all have one. Public *members* of exported classes do not: `Logger.trace`, `.debug`, `.info`, `.warn`, `.startTimer`, `.showUserChannel`, `.showDebugChannel` and `.getBufferSize` lost theirs, because each was the exact restated-signature shape the standard's delete table names (`/** Log a TRACE level message */`), and the rewritten class doc now carries the level semantics they were gesturing at. This matches the pilot, where `ImportHandler`'s four self-evident methods carry no doc.

## Review

Three rounds against a reviewer with a fresh context, briefed to treat "a comment that states something false about the code" as the defect that matters here, since no test catches one. Rounds 1 and 2 both returned FAIL, and every finding was the same species - a comment asserting more than the code supports. Each was verified against the source before being fixed:

| Claim | What the code says | Fixed in |
| --- | --- | --- |
| Handlers "still take an OutputChannel rather than reaching for the singleton" | Inverted: they take the channel and log through `logger` | `28e8e6f` |
| Disposing `StatusBarHandler` stops a snooze "expiring and re-enabling auto-import" | `isSnoozeActive()` is time-based and a snooze never touches the setting; the disposal that matters is its config listener | `28e8e6f` |
| `generatedAt` "restamped every time the data changes" | `handleFileDelete` prunes `nodes` and saves without restamping | `28e8e6f` |
| `UEFNProjectFile` is "the subset this extension reads", every field optional | `plugins[].name` is required, and `projectId` and `plugins` have no reader | `28e8e6f` |
| `findAndParseProjectFile` caches "after the first call" | Only a success is cached | `28e8e6f` |
| The two named `outputChannel` uses are "the only real uses left" | It is handed on at seven sites; the one terminal use is `StatusBarHandler.ts:280` | `eff5743` |
| `generatedAt` is "the age of the last scan" | Invalidation restamps it too | `eff5743` |

Round 3 confirmed all three replacements against the code and returned **PASS** with no findings.

## Verification

No test file was added, removed or edited. No changelog fragment: not a user-facing change.

### Strip-diff gate

Comment-stripped compiled output, base against head:

```
npx tsc -p ./ --removeComments --sourceMap false --outDir out/strip-base   # at origin/main
npx tsc -p ./ --removeComments --sourceMap false --outDir out/strip-head   # at HEAD
diff -r out/strip-base out/strip-head
STRIP-DIFF EXIT: 0
```

Empty. Nothing but comments changed.

### `npm run compile`

```
> tsc -p ./
```

### `npm test`

```
> jest --verbose

Tests:       595 passed, 595 total
Snapshots:   0 total
Time:        7.727 s
Ran all test suites.
```

### `npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -66,8 +66,10 @@ export function activate(context: vscode.ExtensionContext) {
     logger.info("Extension", `Verse Auto Imports ${environment.extensionVersion} is now active (${formatHostSummary(environment, sessionState)})`);
     logger.debug("Extension", "Settings at activation", sessionState.settings);
 
-    // Handlers predate the logger and still take an OutputChannel rather than
-    // reaching for the singleton themselves.
+    // Handlers take this channel but log through the `logger` singleton. The
+    // parameter survives from before the logger existed, and the only real
+    // uses left are StatusBarHandler showing the channel and ProjectPathCache
+    // passing it on to the scanner.
     const outputChannel = logger.getUserChannel();
 
     const config = vscode.workspace.getConfiguration("verseAutoImports");
@@ -177,8 +179,9 @@ export function activate(context: vscode.ExtensionContext) {
         ),
     );
 
-    // Clears the snooze timer on deactivation, so a snooze cannot expire and
-    // re-enable auto-import after the extension is torn down.
+    // Disposal is what removes the handler's own configuration listener, which
+    // would otherwise outlive the extension; it also stops the snooze countdown
+    // and disposes the status bar item.
     context.subscriptions.push(statusBarHandler);
 
     context.subscriptions.push(
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -67,9 +67,9 @@ export function activate(context: vscode.ExtensionContext) {
     logger.debug("Extension", "Settings at activation", sessionState.settings);
 
     // Handlers take this channel but log through the `logger` singleton. The
-    // parameter survives from before the logger existed, and the only real
-    // uses left are StatusBarHandler showing the channel and ProjectPathCache
-    // passing it on to the scanner.
+    // parameter survives from before the logger existed and is threaded down
+    // the whole handler tree; the only thing anyone does with it at the end is
+    // StatusBarHandler showing the channel.
     const outputChannel = logger.getUserChannel();
 
     const config = vscode.workspace.getConfiguration("verseAutoImports");
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -28,9 +28,9 @@ function getExplicitSetting(config: vscode.WorkspaceConfiguration, key: string):
 }
 
 /**
- * Gets the configured debounce delay, handling backward compatibility with
- * the deprecated diagnosticDelay setting. Only explicit overrides are
- * considered so diagnosticDelay's registered default (1000) cannot shadow
+ * The debounce delay in milliseconds, honouring the deprecated
+ * diagnosticDelay setting. Only explicit overrides are considered, so
+ * diagnosticDelay's registered default (1000) cannot shadow
  * autoImportDebounceDelay; an explicit autoImportDebounceDelay wins over an
  * explicit legacy value.
  */
@@ -46,8 +46,14 @@ function getConfiguredDebounceDelay(config: vscode.WorkspaceConfiguration): numb
     return config.get<number>("general.autoImportDebounceDelay", 3000);
 }
 
+/**
+ * Builds the handlers, registers every provider, watcher, command and listener
+ * on the extension context, and starts the two caches.
+ *
+ * Whether the project path cache exists at all is decided once, here, from the
+ * setting as it reads at activation. Nothing downstream can turn it on later.
+ */
 export function activate(context: vscode.ExtensionContext) {
-    // Initialize the logger
     logger.initialize(context);
 
     // Record which build and host this is before anything else logs, so an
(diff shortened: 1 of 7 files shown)


vukefn:
Follow-up for the unread types filed as #254 - covers both `src/types/configuration.ts` and the four dead `moduleInfo.ts` types. Nothing in this PR depends on it.

PR #263, opened 2026-08-09
fix(imports): classify a dotted import by the statement at its head

### LINKED ISSUE #247: A bare dotted import sharing its line is not recognised as an import at all
`ImportFormatter.isModuleImport` keeps its own copy of the dotted pattern, `/^using\.\s+(.+)/`, greedy to end of line. #244 fixed the copy in `matchImport` and deliberately left this one, so the two now disagree about where a dotted statement ends.

The disagreement is only visible where the swallowed content changes the classification, which needs a bare folder-module identifier:

```
in : using. Features; MyVal := 5
```

`isModuleImport` is handed the whole line, captures `Features; MyVal := 5`, and asks its three questions of that: it does not start with `/`, holds no `.`, and does not match `^[A-Za-z_][A-Za-z0-9_]*$`. So the answer is `false`, `scanModuleImports` skips the line whole, and nothing records that the file imports `Features`.

## Why it matters

Nothing is corrupted — a skipped line is never rewritten, which is why this was not folded into #244. The cost is the one `ScannedImport.rebuildLosesText` already names for a different shape: a path that does not count as present lets a writer add a second copy of it. A compiler diagnostic asking for `Features` writes a fresh `using { Features }` above a line that was already importing it.

`allUsingPaths` is unaffected, since it does not classify — it reaches `extractPathFromImport` and now reports `Features` correctly. So the two readers of the same line disagree, which is the hazard `matchImport`'s own doc comment warns about: "a looser second copy that, on the same input, disagrees about which statement was read."

The absolute and dot-notation cases are already safe, because the swallowed content keeps the `/` or the `.` that decides them — `using. /X; MyVal := 5` and `using. Economy.Shop; MyVal := 5` both classify correctly and are pinned by #244.

## Where

`isModuleImport` (`src/imports/ImportFormatter.ts`), the dotted branch. `matchImport` in the same file now holds the terminated pattern as `DOTTED_STATEMENT`, so the fix is likely to route this branch through it rather than to copy the class a second time — which is what would put the two back in step permanently.

Worth deciding at the same time whether `isModuleImport` should answer from the statement at the head of the line at all, given it is asked of whole lines that may carry more than one.

## Suggested acceptance

- `scanModuleImports(["using. Features; MyVal := 5", "", "code()"])` records `Features`, pinned, and organizing leaves the line as written.
- Nothing writes a second `using { Features }` when a diagnostic asks for `Features` and the file holds only that line.
- The existing classification of `using. Instance # see Foo.Bar` as a local-scope using is unchanged.
- Regression tests in `src/imports/__tests__/`.

Found during the review of #244; see PR #246, which scopes it out deliberately.

vukefn:
Closes #247

`isModuleImport` kept its own copy of the dotted pattern, `/^using\.\s+(.+)/`, greedy to end of line, while `matchImport` ends the capture where the path ends. A bare folder-module import sharing its line with a definition was therefore handed `Features; MyVal := 5` to classify: no leading `/`, no `.`, not an identifier, so no import at all. `scanModuleImports` skipped the line whole, nothing recorded that the file imports `Features`, and a diagnostic asking for it wrote a second `using { Features }` above a line already making that import.

The dotted branch now reads `DOTTED_STATEMENT`, so the classifier and the reader answer from the same statement rather than from two opinions about where it ended.

## What changed

- `ImportFormatter.isModuleImport` matches `DOTTED_STATEMENT` in its dotted branch instead of a second, looser copy of it.
- Doc comments carry the two constraints that fall out: `isModuleImport` classifies from the head statement (the indented style excepted, which still reads the whole of `nextLine`), and `DOTTED_STATEMENT` now has two readers, so widening the class widens both.

The ticket's open question - whether `isModuleImport` should answer from the statement at the head of the line at all - is answered yes, and documented on the method.

## Deliberate consequences, beyond the reported symptom

Both follow from sharing the pattern rather than copying it, and both are pinned by tests:

- `DOTTED_STATEMENT` needs no space after the `.`, where the replaced copy required one. `using.Features` now classifies as a module import at file scope. Everything reading through `matchImport` (`allUsingPaths`, `usingPathsOnLine`) already reported a path there, so this closes the same two-readers-disagree gap one form over rather than opening a new one.
- In default mode a dotted local-scope `using` sharing its line no longer gets promoted by a dot in the leftover: `using. Instance; MyVal := Foo.Bar` was a module import and is now correctly a local-scope using.

## Out of scope, needs its own issue

`ImportPathConverter.extractPathFromImport` (`src/imports/ImportPathConverter.ts:122`) holds a third copy of the dotted pattern, `/^using\.\s*(.+)/`, still greedy to end of line. It is safe today only because `scanConvertibleImports` feeds it rewritable single-statement lines, so leftover text never reaches it - an invariant nothing states next to either copy. Not touched here.

## Tests

Regression tests in `src/imports/__tests__/`, each verified to fail against the unfixed regex:

- `ImportScanner`: `scanModuleImports(["using. Features; MyVal := 5", "", "code()"])` records `Features`, pinned - `[]` before.
- `ImportScanner`: a `using:` pair opened after a bare dotted import records both paths, pinned - `[]` before, which stranded the indented path in the body, imported by nothing.
- `ImportDocumentEditor`: organizing with `["Features"]` against that line adds nothing - before, it emitted `using { Features }` above the line, reproducing the reported harm exactly.
- `ImportFormatter`: the three classification flips (bare dotted sharing its line, the no-space form, the local-scope promotion).

Plus a guard that organizing leaves the line as written, and the pre-existing pin on `using. Instance # see Foo.Bar` as a local-scope using, which still passes.

## Review

Fresh-context review gate, `opus`: **PASS_WITH_SUGGESTIONS**, 0 Critical, 0 Important, 4 Suggestions. Three were applied in the second commit (the head-statement doc claim was too broad, one comment narrated the incident the commit already carries, and the default-mode assertion named a promotion it never exercised). The fourth is the third pattern copy, recorded above rather than folded into this diff.

## Verification

```
$ npm run compile

> tsc -p ./

$ npm test

Tests:       602 passed, 602 total
Snapshots:   0 total
Time:        7.433 s, estimated 9 s
Ran all test suites.

$ npm run format:check

Checking formatting...
```

### DIFF
--- a/changelog.d/247.fixed.md
+++ b/changelog.d/247.fixed.md
@@ -1,0 +1,1 @@
+**Import scanning**: A dotted import of a same-directory module sharing its line with another statement now counts as imported, so nothing writes a second copy of an import the file already makes.
--- a/src/imports/ImportFormatter.ts
+++ b/src/imports/ImportFormatter.ts
@@ -90,14 +90,13 @@ export class ImportFormatter {
      *   at file scope a bare `using { X }` can only be the former. Pass this
      *   only when you know the line is not inside a function body.
      *
-     * Every style answers from the statement at the head of the line rather
-     * than from the whole of it, since a line may write more than one and a
-     * definition after a `;` belongs to none of the `using` before it. The
-     * dotted form reads matchImport's DOTTED_STATEMENT for that: its content is
-     * what ends it, and a second copy of the pattern is a second opinion about
-     * where it ended. A bare folder-module name is where the two answers differ:
-     * `Features; MyVal := 5` is neither a path nor an identifier, so a line
-     * read whole classifies as no import and is skipped rather than counted.
+     * The dotted and braced styles answer from the statement at the head of the
+     * line rather than from the whole of it, since a line may write more than
+     * one and a definition after a `;` belongs to none of the `using` before
+     * it. The dotted form reads matchImport's DOTTED_STATEMENT to do that: its
+     * content is what ends it, and a second copy of that pattern is a second
+     * opinion about where it ended. The indented style is the exception and
+     * classifies from the whole of nextLine.
      *
      * @param nextLine The following line, which carries the content for the
      *   indented style. When it is not given and the line is `using:`, this
--- a/src/imports/ImportFormatter.ts
+++ b/src/imports/ImportFormatter.ts
@@ -90,6 +90,15 @@ export class ImportFormatter {
      *   at file scope a bare `using { X }` can only be the former. Pass this
      *   only when you know the line is not inside a function body.
      *
+     * Every style answers from the statement at the head of the line rather
+     * than from the whole of it, since a line may write more than one and a
+     * definition after a `;` belongs to none of the `using` before it. The
+     * dotted form reads matchImport's DOTTED_STATEMENT for that: its content is
+     * what ends it, and a second copy of the pattern is a second opinion about
+     * where it ended. A bare folder-module name is where the two answers differ:
+     * `Features; MyVal := 5` is neither a path nor an identifier, so a line
+     * read whole classifies as no import and is skipped rather than counted.
+     *
      * @param nextLine The following line, which carries the content for the
      *   indented style. When it is not given and the line is `using:`, this
      *   conservatively returns `true`.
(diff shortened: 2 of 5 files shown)


vukefn:
The third copy of the dotted pattern noted above is now tracked as #264. Not touched in this PR.

PR #284, opened 2026-08-10
fix(imports): read a # inside a string or char literal as content

### LINKED ISSUE #272: scanLine reads a # inside a string or char literal as a comment opener
A bare `#` inside a string or char literal is content, not a comment. Compile-validated in the book's suite (verselang/book):

- `assert{" {0o23} " = " # "}` in `Tests/Literals/String.versetest:24`
- `option{'#'}` in `Tests/Literals/Char.versetest:120`

Only `<# ... #>` is special inside strings ("Comments in strings are removed", `docs/01_expressions.md`; `Tests/Literals/String.versetest:234`), and `scanLine` already handles that half correctly.

## Where

`ImportScanner.ts:134-136`: `scanLine` ends its scan at any `#` outside a block comment, with no literal tracking. The design note above it argues string tracking is unnecessary because a `<#` in a string really does open a comment; that argument is right for `<#` and does not transfer to `#`.

## Consequence

Executed on `main` @ 3813ebd:

```
allUsingPaths(['X := "a#b"; using { Economy.Shop }', 'using { /A }', 'code()']) -> ["/A"]
```

The `using` after the literal is live code and is missed; a char literal (`C := '#'; using { Features }`) misses the same way.

Worst case, executed end to end: the miss flips `withholdIsSafe` (`ImportDocumentEditor.ts:911`), organize reads the file as all-absolute and withholds a movable duplicate of a pinned path, removing the provider the hidden bare `using` was resolving against. Every other consequence of the truncation pins lines, which is the safe direction.

## Suggested acceptance

- `scanLine` tracks string and char literals far enough that a `#` inside one stays code, while a `<#` inside a string still opens a comment.
- Regression tests for both literal forms and for the `withholdIsSafe` consequence.

vukefn:
Closes #272

`scanLine` ended its scan at any `#` outside a block comment, with no literal tracking, so a `#` written inside a string or char literal truncated the line and hid every statement after it. A live `using` behind one went uncounted, which can flip `withholdIsSafe` and let organize remove an import the hidden `using` was resolving against.

Literals are now tracked far enough to tell content from a comment opener:

- `<#` keeps first refusal, because a `<# ... #>` written inside a string really is a comment there (`"abc<#def#>ghi"` is `"abcghi"`), and the literal survives across it — `"a<#c#>#b"` is the string `"a#b"`.
- Backslash escapes are consumed, so `"a\"b#c"` and `'\''` do not close early.
- A line still inside a literal at its end is read again with literals ignored, which is byte-identical to the previous behaviour. A string may legitimately be open at end of line, since an interpolation block spans lines, and nothing can be lexed past that point — treating the remainder as literal content is what would let a `using` written in a trailing comment read as the line's own statement. This bounds the change to lines whose literals balance, so no input that behaves correctly today can regress.

Verified against the language's own suite in `verselang/book`: `assert{" {0o23} " = " # "}` (`Tests/Literals/String.versetest`), `option{'#'}` (`Tests/Literals/Char.versetest`), and `assert{"abc<#def#>ghi" = "abcghi"}` for the `<#` half.

## Known limit, unchanged by this PR

A quote inside a string's `{ ... }` interpolation ends the literal here, so `X := "a{F("#")}b"; using { /A }` still misses the `using`. That is the same miss narrowed rather than closed; `main` behaves identically, and interpolation-aware lexing is beyond what the ticket asks. Recorded in the function's doc comment.

## Tests

Seven new assertions, each proved to fail against the unfixed `scanLine` by stashing it:

```
  ● allUsingPaths › collects a using written after a char literal holding a #
  ● allUsingPaths › collects a using written after a string literal holding a #
  ● allUsingPaths › does not end a string literal at an escaped quote
  ● classifyLines › keeps a # inside a char literal in the line's code
  ● classifyLines › keeps a # inside a string literal in the line's code
  ● classifyLines › keeps a string open across a block comment written inside it
  ● ImportDocumentEditor.buildOrganizedContent › keeps a duplicate when a using behind a string literal could resolve against it
Tests:       7 failed, 265 passed, 272 total
```

Two further assertions pin the halves that were already right — a `<#` inside a string still opening a comment, and an unterminated literal keeping the old reading — and pass in both directions.

Review gate: PASS_WITH_SUGGESTIONS, 0 Critical, 0 Important, 5 Suggestions. Four were applied: the marker exception is narrowed to `<#` (the shipped grammar starts a block comment at `<#(?!>)`, so `<#>` is not one and its precedence inside a string is a separate question), the interpolation limit is documented, a redundant `hasCode` assignment in the escape branch is removed, and two comments explaining the fallback by reference to the previous implementation are rewritten as the constraint itself.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Test Suites: 29 passed, 29 total
Tests:       640 passed, 640 total
Snapshots:   0 total
Time:        7.731 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/272.fixed.md
+++ b/changelog.d/272.fixed.md
@@ -1,0 +1,1 @@
+**Import scanning**: A `#` inside a string or char literal no longer ends the scan of its line, so a `using` written after one is seen and organize keeps an import it would otherwise remove.
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -104,18 +104,18 @@ interface LineScan {
  * - A `#` line comment runs to the end of the line, so a `<#` inside one is
  *   text rather than an opener.
  * - A bare `#` inside a string or char literal is content, so a line is lexed
- *   far enough to tell the two apart. `<# ... #>` is the exception in both
- *   directions: it really does open a comment inside a string, so the markers
- *   are read before the literal state is consulted, and a literal survives
- *   across the comment written into it.
+ *   far enough to tell the two apart, and no further: a quote inside a string's
+ *   `{ ... }` interpolation ends the literal here, so a `#` after one still
+ *   reads as an opener. `<#` is the exception in both directions - it really
+ *   does open a comment inside a string, so it is read before the literal state
+ *   is consulted, and a literal survives across the comment written into it.
  *
- * A line that ends inside a literal is read again with literals ignored, which
- * is how this read every line before they were tracked at all. Nothing was
- * lexed past the point the literal was left open - a string may legitimately be
- * open there, since an interpolation block spans lines - so the rest of the
- * line is trivia this cannot classify, and treating it as literal content is
- * what would let a `using` written in a trailing comment read as the line's own
- * statement.
+ * A line still inside a literal at its end is read again with literals ignored.
+ * Nothing could be lexed past the point the literal was left open - a string may
+ * legitimately be open there, since an interpolation block spans lines - so the
+ * rest of that line is trivia this cannot classify, and treating it as literal
+ * content is what would let a `using` written in a trailing comment read as the
+ * line's own statement.
  */
 function scanLine(line: string, depth: number): LineScan {
     return lexLine(line, depth, true) ?? lexLine(line, depth, false)!;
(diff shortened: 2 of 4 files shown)


vukefn:
The residual gap this PR documents but does not close is now tracked as #285: a quote inside a string's `{ ... }` interpolation still ends the literal, so `X := "a{F("#")}b"; using { /A }` is missed. Executed against this branch: `code` truncates to `X := "a{F("` and `allUsingPaths` returns `[]`, where the plain case now returns `["/A"]`.

PR #291, opened 2026-08-10
fix(services): accept every module declaration form in the project scan

### LINKED ISSUE #282: ProjectPathScanner's module pattern accepts only the colon form, so brace and dotted declarations are missed
`ProjectPathScanner.scanFile` classifies a declaration line with its own table of patterns, and the module entry (`src/services/ProjectPathScanner.ts:162`) accepts only the colon form:

```ts
const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*:/;
```

Verse writes a macro body in three equivalent styles, and `module` takes all of them. So a module declared either of these two ways is not recognised as a module by this scan:

```verse
Inventory := module { }

Inventory := module. Item := module{}
```

The brace form with the brace on the following line is missed as well, and so is the `>` terminator the sibling pattern in `ImportPathConverter` accepts.

## Why it is a third copy

`ImportPathConverter.buildModuleDefinitionRegex` (`src/imports/ImportPathConverter.ts:109`) already ends at `[:>{.]`, which covers all four. #281 widened it there and deliberately left this one alone, on the grounds that the declaration cache is only the second of `findModuleLocations`' three phases and the explicit search it fixed is the third, so the module still resolves either way. That reasoning holds for path conversion. It does not cover the other readers of the cache.

#260 has since removed `ImportPathConverter`'s second copy, so this is now the one remaining site in `src/` that spells the module-declaration grammar independently of `buildModuleDefinitionRegex`.

## Why it is worth its own issue rather than folding into #45

#45 plans one shared declaration parser across all five consumers and names the project tree builder among them, so this is inside its eventual scope. But #45 is a pre-0.7.0 PRD covering the digest build script, the runtime and assets digest parsers, and the type and manifest sharing alongside the parsing — a large piece of work. This is one character class in one line, and it is a live blind spot in the cache's view of the project until that lands.

The narrower question here is also different from #45's: whether `modulePattern` should widen in place, or whether this table's module entry should call `buildModuleDefinitionRegex` with a wildcard. The second is a step toward #45; the first is a stopgap.

## Constraint on any fix

The comment above the pattern table (`ProjectPathScanner.ts:156-161`) records why `modulePattern` alone allows no specifiers to the right of `:=` — specifiers there belong to the type, not the name, and the other patterns skip rather than capture them. Whatever replaces it has to keep the two capture groups the loop below reads: `[1]` the declared name, `[2]` the specifiers attached to it.

## Acceptance

- A module declared in the brace form, the brace-on-next-line form or the dotted form is recorded as a module by `ProjectPathScanner`, not skipped or mistyped.
- A test per form beside the existing scanner tests, each confirmed to fail against the unfixed pattern.
- No change to which declarations the other patterns in the table accept.

Deferred from #281, found again while closing #260.

vukefn:
Closes #282

`ProjectPathScanner`'s module row ended at `module\s*:`, so only the colon form was recorded. The other forms were not merely mistyped — the fall-through `variablePattern` matched the `:` of `:=`, and its guard then skipped the line for naming `module`, so the declaration left no node at all and every reader of the cache was blind to it.

```diff
-const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*:/;
+const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*(?:[:>{.]|$)/;
```

`[:>{.]` is the terminator set `ImportPathConverter.buildModuleDefinitionRegex` already accepts. The `|$` alternative is what this site needs and the siblings do not: they match whole file text, so their `\s*` crosses the newline to a next-line brace, while this scan sees one trimmed line at a time. Both capture groups are unchanged — `[1]` the name, `[2]` the specifiers attached to it — which is the constraint the issue sets.

## On the approach

Widened in place, as the issue's stopgap option. The alternative it names — calling `buildModuleDefinitionRegex` with a wildcard — is not viable as written: that function escapes backslashes in its argument, captures neither the name nor the specifier block, and anchors on `\b` rather than line start.

One correction to the issue's premise: it says this is "the one remaining site in `src/` that spells the module-declaration grammar independently". #286 landed `MODULE_DECLARATION` in `src/visibility/moduleDeclarations.ts` after the issue was filed, so there are now three spellings, not two. Sharing that one here would still mean composing rather than reusing (the `|$` again), would invert the layering, and would silently widen the name grammar to quoted suffixes. Unifying them stays #45's job; the comment above the pattern now names both siblings so the triplication is findable when it lands.

## Known limits, unchanged by this PR

- A trailing comment after the keyword (`X := module # opens below`) still defeats the match. `buildModuleDefinitionRegex` has the identical gap.
- The scanner masks no comments, so a brace-form module inside a `<# ... #>` block is now recorded as one — the same pre-existing hole that already swallowed the colon form.
- Members of a next-line-brace module are not nested: the bare `{` sits at the declaration's own indent and indentation alone closes a module. Pinned by a test rather than left to be discovered.

## Tests

`src/services/__tests__/ProjectPathScanner.test.ts` is new — the class had no test file. One test per form; 6 of the 7 form tests fail against the unfixed pattern (`git stash push -- src/services/ProjectPathScanner.ts`), the seventh being the `modulex` negative, which must hold in both directions.

## Review

Fresh-context reviewer, opus: **PASS_WITH_SUGGESTIONS** — 0 Critical, 2 Important, 4 Suggestions. Both Important findings are applied here: the missing changelog fragment, and the now-false sentence in `moduleDeclarations.ts` describing this scanner as colon-only.

## Verification

```
> tsc -p ./

Test Suites: 38 passed, 38 total
Tests:       745 passed, 745 total
Snapshots:   0 total
Time:        229.001 s
Ran all test suites.

> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/282.fixed.md
+++ b/changelog.d/282.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: a module declared as `Name := module { ... }`, with the brace on the next line, or as `Name := module. ...` now enters the project's declaration cache, so module lookups and quick fixes see it.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -160,12 +160,12 @@ export class ProjectPathScanner {
         // rather than captured - except by modulePattern, which allows none
         // there and so does not match a module carrying them.
         //
-        // A module body is written in any of the three styles Verse gives a
-        // macro, so the keyword ends at `:`, `{` or `.`, and at the line end
-        // when the brace opens on the next line. `>` matches the terminators
-        // ImportPathConverter.buildModuleDefinitionRegex accepts. This line is
-        // matched trimmed, so the alternative is what reaches a next-line
-        // brace that regex reads through `\s*` over the whole file.
+        // A module body takes any of the three styles Verse gives a macro, so
+        // the keyword ends at `:`, `{` or `.` - or at the line end, which is
+        // how a next-line brace reaches a scan that sees one line at a time.
+        // `>` holds parity with the two other spellings of this grammar,
+        // ImportPathConverter.buildModuleDefinitionRegex and
+        // moduleDeclarations' MODULE_DECLARATION.
         const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*(?:[:>{.]|$)/;
         const classPattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*class\s*(?:<[^>]+>)*\s*[\(:]?/;
         const structPattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*struct\s*(?:<[^>]+>)*\s*[\(:]?/;
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -159,7 +159,14 @@ export class ProjectPathScanner {
         // right of `:=` belong to the type, not the name, and are skipped
         // rather than captured - except by modulePattern, which allows none
         // there and so does not match a module carrying them.
-        const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*:/;
+        //
+        // A module body is written in any of the three styles Verse gives a
+        // macro, so the keyword ends at `:`, `{` or `.`, and at the line end
+        // when the brace opens on the next line. `>` matches the terminators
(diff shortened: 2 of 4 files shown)


vukefn:
Follow-up filed from the review gate: #292 — this scanner masks no comments, so a declaration inside `<# ... #>` is recorded as live. Pre-existing (the colon form was already recorded), widened here to the brace and dotted forms.

Issue #270, opened 2026-08-09
Placement bounds ignore imports pinned by their span, so a new dotted import lands above its pinned provider


vukefn:
Under default settings (`behavior.preserveImportLocations: true`, sorting on), a new path is merged into an existing block bounded by the anchored imports. Only imports flagged `anchorsCommentBelow` are collected as bounds:

```ts
// ImportDocumentEditor.ts:501
const anchoredImports = scannedImports.filter((imp) => imp.anchorsCommentBelow);
```

An import pinned by `rebuildLosesText` alone (a line that carries another statement) is excluded from `rewritableImports`, so it is in no block, and it is not in `anchoredImports`, so it raises no floor and no ceiling. Every placement decision is blind to it.

## Reproduction

Executed through the `fakeDocument` harness on `main` @ 3813ebd. Input:

```
using { /Verse.org/Simulation }
X := 1
using { Features }; MyVal := 5
code()
```

Adding `using { Economy.Shop }` produces one replace at lines 0-1:

```
using { /Verse.org/Simulation }
using { Economy.Shop }
```

The new dotted import now sits above the pinned `using { Features }` on line 2. Where `Economy` is a member of `Features`, the import cannot resolve at the line it was written on, while a legal position below the pinned provider exists; the "Unknown identifier" the import was added to fix survives.

## Where

- `ImportDocumentEditor.ts:501` builds the bounds list from `anchorsCommentBelow` only.
- `anchoredBounds`, `blockIsWithin` and the `selectTargetBlockIndex` eligibility filter all read that list, so every placement site inherits the blindness.

## Context

#196 fixed this class for anchored imports. #232 covers the organize path for both pin reasons, #221 the hoist-by-contract paths. This is the remaining quadrant: the add path under `preserveImportLocations: true` with a pinned-by-span provider.

## Suggested acceptance

- An import pinned for either reason participates in the same placement bounds an anchored one does.
- Regression tests in `src/imports/__tests__/`, running without the VS Code runtime, covering a pinned bare provider below the first block and a newly added dotted consumer.

Issue #221, opened 2026-08-08
Hoisting imports to the top can place one above the anchored import that provides it


vukefn:
Hoisting imports to the top can place one above the anchored import that provides it.

#196 fixed this for the *add* path under `behavior.preserveImportLocations: true`, where placement is a decision. Two paths hoist to the top by contract instead, and neither consults the anchored imports:

- `ImportDocumentEditor.addImportsToDocument`, the `preserveImportLocations: false` branch. It writes every relocatable path plus the new ones at `Position(0, 0)` and deletes the movable blocks. An anchored import stays where it is, so a new dotted consumer lands above the anchored bare provider that brings its first segment into scope.
- `ImportDocumentEditor.buildOrganizedContent`, behind the organize command. It rebuilds the block after the file header and leaves the anchored line in the body below, so a *live* movable consumer hoisted into that block ends up above an anchored provider that was already beneath it.

## Why this was not folded into #196

It is a contract question rather than a placement bug. Both paths promise "all imports at the top", and that promise is what conflicts with an anchored import sitting lower - the fix is to decide what the contract becomes, not to add a bound. `buildOrganizedContent` already reasons about anchored imports for a neighbouring question (whether it is safe to withhold a duplicate path, `ImportDocumentEditor.ts` around the `withholdIsSafe` check) and deliberately says nothing about placement.

Changing one without the other would also make the two diverge on where the block goes, which is worse than the present state: today they agree.

## Reproduction

```
using { Features } <#> why this is here
    it brings Economy into scope

code()
```

With `behavior.preserveImportLocations: false`, a diagnostic asking for `Economy.Shop` writes `using { Economy.Shop }` at line 0, above the `using { Features }` that provides `Economy`. Verse resolves `using` sequentially, so the "Unknown identifier" the import was added to fix survives.

For the organize path, start from a file where a live `using { Economy.Shop }` sits below an anchored `using { Features }` and run Organize Imports: the consumer is hoisted above the provider.

## Confidence

Read from the code paths, same as #196's original diagnosis. Not executed end to end. The add-path half is unit-testable through the `fakeDocument` harness in `src/imports/__tests__/ImportDocumentEditor.test.ts`; `buildOrganizedContent` is a pure string function and needs no harness at all.

## Suggested acceptance

- Decide the contract first: does a hoisted block move below a constraining anchored import, or does the anchored import's provided path get withheld from the block, or is the file left alone with a warning?
[...]

Issue #196, opened 2026-08-07
A new import can be placed above the anchored import that provides it


vukefn:
An import anchored by a comment marker belongs to no import block, so the placement logic added for #129 cannot see it, and a new import can be inserted above the anchored import that provides its first segment. The "Unknown identifier" the import was added to fix then persists.

## Where it comes from

`addImportsToDocument` builds its blocks from the movable imports only:

```ts
// ImportDocumentEditor.ts:300
for (const imp of rewritableImports(scannedImports)) {
```

`rewritableImports` drops everything flagged `anchorsCommentBelow`, so an anchored import is in no `ImportBlock`. Every placement decision downstream reads `importBlocks`:

- `selectTargetBlockIndex` (`ImportDocumentEditor.ts:234-253`) ranks the new path against `block.imports` only. Its own doc comment states the job it cannot do here: *"the first block holding an import the new one belongs above: a higher rank. This is what keeps a bare provider above the dotted consumer it was added for"* (`:224-226`). An anchored provider is in no block, so no bound is ever derived from it.
- With the anchored import as the file's only import, `importBlocks` is empty, so the `sortAlphabetically && importBlocks.length > 0` branch at `:438` is skipped and `:499` inserts at `Position(0, 0)` — the top of the file, above the anchored line.

## Expected

```
using { Features } <#> why this is here
    it brings Economy into scope
code()
```

A diagnostic asking for `Economy.Shop` should place `using { Economy.Shop }` **below** the anchored `using { Features }`, since Verse resolves a `using` top-down and `Economy.Shop` needs `Economy` in scope above it.

## Actual

`using { Economy.Shop }` is inserted at line 0, above `using { Features }`. Same breakage as #129 and #91, reached through anchoring rather than through blocks.

## Confidence

The mechanism is read from the code paths above and is unambiguous — `rewritableImports` at `:300` is what removes anchored imports from consideration, and `:499` is the no-blocks insertion point. **The end-to-end reproduction has not been executed.** `addImportsToDocument` needs a real `TextDocument` and `WorkspaceEdit` and is mocked in every existing test, and `selectTargetBlockIndex` is private, so there is no cheap unit path; confirming it wants the Extension Development Host, or a new seam.

## Scope

[...]

PR #297, opened 2026-08-10
fix(imports): bound the organized block against the imports pinned below it

### LINKED ISSUE #232: Organizing hoists an import above the pinned line that provides its first segment
An import whose line may not be rebuilt stays where it was written, and the rest of the file is organized around it. Nothing checks whether the rebuilt block above it now holds a path that resolves against it:

```
in : using { Features } <# note
     using { Old }
     #>
     using { Economy.Other }

     code()

out: using { Economy.Other }

     using { Features } <# note
     using { Old }
     #>

     code()
```

`Economy.Other` is a rank-1 path: its first segment has to be in scope above it. `Features` is what brings `Economy` into scope, and it is now below. The file stops compiling.

## Where

`buildOrganizedContent` (`src/imports/ImportDocumentEditor.ts:585`) splits the scan into the imports it may move and the ones it may not:

```ts
const allImports = scanModuleImports(lines);
const movableImports = rewritableImports(allImports);
```

`movableImports` is sorted and emitted at the top of the file, `headerLineCount` aside. A pinned line is left in the body wherever it sits. The sort that ranks paths so a provider precedes its consumer - `ImportFormatter.sortImportsByRank`, and the defects behind it - only ever sees `movableImports`, so a provider that is pinned is outside the ordering it is supposed to participate in.

Adding a path is the same: `buildOrganizedContent(input, ["Economy.Other"], ...)` writes it into the block above the pinned provider.

The neighbouring `withholdIsSafe` check (`:632`) is the one place that already reasons about this hazard, but it answers a narrower question - whether a *duplicate* of a pinned path may be withheld - and it asks only about the whole file's ranks, not about ordering.

## Scope

Both reasons a line can be pinned reach it, and neither is new:

- `anchorsCommentBelow`, since #166 - the shape above, reproducible on `main` today.
- `rebuildLosesText`, added by #219 for a `using:` pair opened after a `;`. `["using { Features }; using:", "    Economy.Shop", "using { Economy.Other }", "code()"]` produces the same inversion.

#219 did not introduce it and deliberately did not widen into it. Found by that PR's review gate, which reproduced both shapes.

## Suggested acceptance

- Organizing a file whose pinned import provides the first segment of a path in the rebuilt block does not place that path above the pinned line.
- A file with no pinned import is unaffected, and the existing sort is unchanged.
- Regression tests in `src/imports/__tests__/`, covering both pin reasons, which run without the VS Code runtime.

## The open question

What to do when the ordering cannot be satisfied - a pinned line halfway down the file with several consumers above it. Declining to hoist the affected paths at all is the conservative answer, and matches how `withholdIsSafe` errs: leave the file untidy rather than uncompilable.

vukefn:
Closes #232

`buildOrganizedContent` hoisted every rewritable import into one block at the top and left each pinned line in the body, without ever asking whether a pinned import brings the first segment of a hoisted path into scope. A `using` resolves its own path top-down, so the rebuilt file stopped compiling.

## What changed

Nothing is carried across a pinned import that has to precede it.

- **An import already written above such a line is hoisted as before.** The block goes above that line too, so its relation to the pinned import is unchanged and the sort still repairs the order. Holding these back would leave a file the sort could have fixed unrepaired.
- **One written below it keeps its line** — not hoisted, not rebuilt, its comments left with it. Untidy rather than uncompilable, which is how the neighbouring `withholdIsSafe` errs and the answer the issue proposed to its own open question.
- **An additional path has no line to keep**, so it is written directly below the pinned statement that constrains it. Declining to write it would make "organize, then add the missing imports" silently add nothing.

The bound is the floor `pinnedBounds` already computes for the add path (#196, #270), from the same `belongsAbove`, so the two writers read one rule rather than two — the divergence #221 warns about. No ceiling is needed: the block precedes every pinned line already, and so does a path written directly below the floor.

Both pin reasons are covered — `anchorsCommentBelow` (under both markers) and `rebuildLosesText`.

## Acceptance

- Organizing a file whose pinned import provides the first segment of a path in the rebuilt block does not place that path above the pinned line — `src/imports/__tests__/ImportDocumentEditor.test.ts`, five cases including the issue's own reproduction.
- A file with no pinned import is unaffected and the sort is unchanged — pinned empty means `crossesPinnedProvider` is always false, so `hoistableImports === movableImports`; pinned by a test.
- Regression tests live in `src/imports/__tests__/` and run without the VS Code runtime.

One existing expectation changed: "keeps a duplicate when a dotted import could resolve against it". Its `Economy.Shop` sits below the anchored line, so it now stays there; the duplicate the test was written to pin is still kept, which is what it asserts.

Each new test was run against `origin/main` and fails there — 6 of the 9 in the group, the other three being the guards that must hold before and after.

## Not in scope

`addImportsToDocument`'s `preserveImportLocations: false` branch, the other half of #221, which is untouched and still open.

`placementLine`'s doc calls the rank-0-above-rank-1 floor a "preference, raised by a pinned absolute path that brings nothing into scope". An absolute import does bring names into scope — it exposes the public member modules of what it imports under their bare names, which is why a pinned `using { /A }` genuinely constrains a bare reference below it, as the new test asserts. That sentence is #270's and reopening its ceiling-wins reasoning belongs with #221, so it is left alone here and flagged rather than edited.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
> jest --verbose

Test Suites: 37 passed, 37 total
Tests:       746 passed, 746 total
Snapshots:   0 total
Time:        7.84 s

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

Two review rounds, both `PASS_WITH_SUGGESTIONS`, four Important findings between them, all fixed:

1. Grounding ignored position, holding back imports already above the pinned line and losing a repair the sort used to make.
2. `extraPaths` was no longer deduplicated on the branch that bypasses the block's set, so a path asked for twice was written twice.
3. The same branch also missed a path an import in the block already carried, which emitted it in both places — a regression introduced by the fix for (1), now pinned by its own test.
4. The method doc still stated the superseded rule after (1) landed.

### DIFF
--- a/changelog.d/232.fixed.md
+++ b/changelog.d/232.fixed.md
@@ -1,0 +1,1 @@
+**Organize Imports**: an import is no longer hoisted above the import pinned to its line that brings its first segment into scope, so organizing a file holding one leaves it compiling
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -351,6 +351,34 @@ export class ImportDocumentEditor {
     }
 
     /**
+     * The last line a pinned import forbids `path` from being hoisted above, or
+     * -1 where none does.
+     *
+     * A rebuilt block is written above every pinned line, so a pinned import
+     * that has to precede `path` rules the block out for it altogether. The
+     * floor is therefore read as a yes or no first, and only then as the line a
+     * path with none of its own is written below.
+     *
+     * Taken from pinnedBounds so that the writer that rebuilds a block and the
+     * writer that adds a single import cannot disagree about what precedes
+     * what: the same floor, from the same belongsAbove. The ceiling that comes
+     * with it says nothing here - a block above every pinned line already
+     * satisfies one, and a path written directly below this floor is above
+     * every pinned import further down.
+     *
+     * A pinned import never bounds its own path. Nothing needs itself in scope,
+     * and the duplicate of a pinned path the withhold rules deliberately keep
+     * would otherwise be stranded in the body instead of being organized.
+     */
+    private hoistFloor(pinned: ScannedImport[], classifications: LineClassification[], path: string): number {
+        return this.pinnedBounds(
+            pinned.filter((imp) => imp.path !== path),
+            classifications,
+            path,
+        ).floor;
+    }
+
+    /**
      * Where the imports pinned to their line allow a new `path` to be written.
      * See PinnedBounds.
      */
@@ -849,6 +877,12 @@ export class ImportDocumentEditor {
      * first segment against the copy being removed; see the withholdIsSafe
      * check.
      *
+     * A pinned import also bounds the block the file is organized around: a
+     * path it may have to precede is not hoisted above it. An import already
+     * written keeps its line, untidy but compiling; an additional path, which
+     * has no line to keep, is written directly below the pinned statement that
+     * constrains it. See hoistFloor.
+     *
      * The result keeps the text's own line ending, so rebuilding an already
      * organized document reproduces it byte for byte and organizeImports can
      * skip the edit.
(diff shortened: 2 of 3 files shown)


vukefn:
Filed the `placementLine` doc inaccuracy noted above as #298 rather than folding it in: the sentence is load-bearing for #270's ceiling-wins rule, so correcting it is contract reasoning that belongs with #221.

PR #308, opened 2026-08-10
fix(imports): count a using written after a definition on its line as present

### LINKED ISSUE #273: A using written after a non-using definition on its line never counts as present
Confirmed legal by a live compiler check (2026-08-10): a file-scope line such as

```
X := 1; using { /Verse.org/Random }
```

compiles; `;` separates statements exactly as a newline does.

`scanModuleImports` requires a line to begin with `using` (the `isModuleImport` gate, `ImportScanner.ts:321`) and skips this line whole, deliberately (`ImportScanner.ts:379-381`). Presence in `addImportsToDocument` reads that scan (`ImportDocumentEditor.ts:520`), so `/Verse.org/Random` never counts as present and a diagnostic asking for it writes a second copy.

Executed on `main` @ 3813ebd:

```
scanModuleImports(["X := 1; using { /Verse.org/Random }", "code()"]) -> []
allUsingPaths(["X := 1; using { /Verse.org/Random }", "code()"])   -> ["/Verse.org/Random"]
```

`allUsingPaths` does see it, so removal safety is unaffected, and the line itself is never touched. The cost is a duplicate import.

## Suggested acceptance

- The path of a `using` written after a non-using statement counts as present for deduplication, while its line stays pinned exactly as today.

vukefn:
Closes #273

## What was wrong

`scanModuleImports` classifies a candidate line from its head, so `ImportFormatter.isModuleImport` rejects any line whose `using` follows a definition — legal Verse, since `;` separates definitions in a scope exactly as a newline does. The line was skipped whole, nothing recorded that the file imports the path, and a diagnostic asking for it wrote a second copy above a line already making that import.

Reproduced on `d2a1413`:

```
scanModuleImports(["X := 1; using { /Verse.org/Random }", "code()"]) -> []
allUsingPaths(["X := 1; using { /Verse.org/Random }", "code()"])     -> ["/Verse.org/Random"]
```

`allUsingPaths` saw it throughout, so removal safety was never affected and the line itself was never touched. The cost was the duplicate.

## The fix

Three commits.

**`49ee7de`** records every complete `using` statement such a line writes, pinned. Presence is judged from every scanned import including pinned ones, while the rebuilt block is built only from `rewritableImports`, so the path stops being written twice and no writer may move or rebuild the line. The classification gate is left refusing these lines rather than widened, so everything below it may keep testing the raw line, as the `using:` branch documents it must.

**`371775c`** closes a data-loss defect the first commit made common. The lexer kept literal text in the `code` it reports, so a `using` written as string data was offered as a statement the line makes. A path that counts as present is withheld from the rebuilt block, on the grounds that the line holding it already makes the import — so organizing a file whose string happened to name an imported path *deleted* the real `using`:

```
using { /A }                     using { /B }
using { /B }                ->
Hint := "using { /A }"           Hint := "using { /A }"
code()                           code()
```

Literal contents are now masked into `codeOutsideLiterals`, and every site whose paths count as imported reads that. `code` still answers what text a line holds, which is a question about the line rather than about the statements it writes.

That defect was reachable before this branch through a line that was itself an import — `using { /A }` + `using { /B }; Hint := "using { /A }"` loses `/A` on `d2a1413` too — so both shapes are covered and both are regression-tested. `allUsingPaths` deliberately keeps reading unmasked `code`, preserving the over-reporting its only consumer, `withholdIsSafe`, is documented to depend on.

**`6c54f73`** confines the mask to `"` strings. A `'` opens a char literal in the lexer, but the same character closes a path's quoted segment suffix, so masking it truncated `Economy.Shop'Loc'` to `Economy.Shop` — the exact misreading `DOTTED_STATEMENT`'s doc exists to prevent, since that is a different module that exists. Both harms returned through it, including deleting a real import. Nothing is given up by exempting `'`: a char literal holds one character, and a `using` does not fit in one.

Two tests written in the second commit were deleted rather than adjusted here. Neither pinned a shape Verse compiles — `'using { /A }'` is not a valid char literal, and a `using` is a definition rather than an expression, so it cannot appear in a string interpolation.

## Deliberately not covered

`X := 1; using:` opening an indented pair stays skipped whole, as it is today. Consuming it means reasoning about the pair's span, which is a separate change; the doc comment now says so rather than implying deduplication reaches it.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...

$ npm test
> jest --verbose

Test Suites: 39 passed, 39 total
Tests:       797 passed, 797 total
Snapshots:   0 total
Time:        13.302 s
Ran all test suites.
```

Each new test was proven to detect its defect before the fix went in: 4 failures for the deduplication cases with the first commit's change removed, and every data-loss case above was reproduced directly against the commit that introduced it.

Both later commits answer a blocking finding from an independent reviewer given the diff with no knowledge of how it was written.

### DIFF
--- a/changelog.d/273.fixed.md
+++ b/changelog.d/273.fixed.md
@@ -1,1 +1,1 @@
-**Imports**: A `using` written after another statement on the same line now counts as already imported, so a quick fix no longer adds a second copy of it.
+**Import scanning**: A `using` written after another statement on the same line now counts as already imported, and a `using` written inside a string is no longer read as one at all.
--- a/changelog.d/273.fixed.md
+++ b/changelog.d/273.fixed.md
@@ -1,0 +1,1 @@
+**Imports**: A `using` written after another statement on the same line now counts as already imported, so a quick fix no longer adds a second copy of it.
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -84,24 +84,32 @@ interface LineScan {
      */
     code: string;
     /**
-     * `code` with the contents of every string and char literal replaced by
-     * spaces, so a `using` written as string text is not offered as a statement
-     * the line makes.
+     * `code` with the contents of every `"` string replaced by spaces, so a
+     * `using` written as string text is not offered as a statement the line
+     * makes.
      *
      * Read this wherever a path is going to count as imported. The text of a
-     * literal is data, not a statement, and recording a path from it both
+     * string is data, not a statement, and recording a path from it both
      * suppresses an import the file needs and - because a path that counts as
      * pinned withholds the movable import of the same path from the rebuilt
      * block - deletes the real `using` the author wrote.
      *
-     * Spaces rather than nothing, so the text on either side of a literal does
+     * A `'` is deliberately not masked, though the lexer opens a char literal
+     * on one. The same character closes a path's quoted segment suffix,
+     * `Economy.Shop'Loc'`, and masking it truncates the path to `Economy.Shop`,
+     * a different module that exists; see ImportFormatter.DOTTED_STATEMENT.
+     * Nothing is lost by that, because a char literal holds one character and a
+     * `using` cannot fit in one.
+     *
+     * Spaces rather than nothing, so the text on either side of a string does
      * not join into a token neither side wrote. `code` removes comments without
      * replacement on purpose, for the opposite reason; the two differ because a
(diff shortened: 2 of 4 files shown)


vukefn:
The sibling shape, a `using:` pair opened after a non-using definition, is out of scope here and filed as #309. The `scanModuleImports` doc comment in this PR states the exclusion rather than implying deduplication reaches it.

PR #312, opened 2026-08-11
fix(imports): keep relative imports in written order, whichever form they take

### LINKED ISSUE #271: The rank sort assumes only bare imports provide scope, so a dotted provider sorts below its bare consumer
Needs live UEFN validation before fix design, like #65 did: the failing chain is derived from the book, and no compile-validated test pins it.

## The model

- `ImportFormatter.sortImportsByRank` (`ImportFormatter.ts:335-349`): rank 0 (absolute) before rank 1 (bare) before rank 2 (dotted); rank 2 alphabetized; only rank 1 keeps written order.
- `ImportDocumentEditor.couldResolveAgainst` (`ImportDocumentEditor.ts:346-348`): only `existing === 2 && rank < 2` raises a ceiling, so only a dotted import is ever treated as a consumer.

The model encodes "a dotted import may depend on a bare one, never the reverse".

## The language

Imports resolve sequentially top-down, and an import of a scope makes that scope's public member-module names available to later imports:

- absolute provider before bare consumer: verselang/book `Tests/Modules/Using2.versetest` M1 (`assert_valid`)
- bare provider before bare consumer: verselang/book `docs/16_modules.md`, the Nested Imports versetest

By the same mechanism, `using { GameSystems.Inventory }` should expose Inventory's member modules, so a later bare `using { Weapons }` (Weapons nested in Inventory) resolves against the dotted import. That spelling is the one no book test pins.

## What the sort does

Executed on `main` @ 3813ebd:

```
sortImportsByRank(["GameSystems.Inventory", "Weapons"]) -> ["Weapons", "GameSystems.Inventory"]
```

Under default settings, organizing a file holding that pair inverts a compiling order. Rank-2 alphabetization has the same defect between two dotted imports where the first provides the second's head (`using { Systems.Registry }` then `using { Economy.Shop }`).

`ImportFormatter.test.ts:237` pins the current rule and becomes a wrong-rule test if validation confirms.

## Validation case

In a live UEFN project: a module `Weapons` nested inside `GameSystems/Inventory`; a file with `using { GameSystems.Inventory }` then `using { Weapons }`. If it compiles and the reversed order fails, the finding is real. If the compiler refuses the chain, this is a wontfix and the model is exact.

## Suggested acceptance (if validated)

- The sort never moves an import above another that can provide for it; written order is the fallback wherever a provider relationship is possible, as it already is for bare-bare pairs.

vukefn:
Closes #271

A `using` of any scope exposes that scope's public member modules to the statements below it, so a dotted import can bring into scope the module a later bare import names — `using { GameSystems.Inventory }` then `using { Weapons }`. The rank model read that relationship one way only, so `sortImportsByRank` sank a dotted provider below its bare consumer and block selection wrote a newly added import above relative imports already written above it.

## The rule now

Only an absolute path resolves from the global registry and needs nothing above it. Every relative path, dotted and bare alike, resolves its first segment against what a `using` above it brought into scope, so either can provide for the other and written order is the only evidence of which way round it is.

- Absolute paths still hoist to the top and alphabetize among themselves.
- Every relative import keeps its written order.
- A newly added relative import goes last, below everything the file already holds.

`ImportFormatter.resolvesAgainstScopeAbove` is the single name for that question. `belongsAbove` and `belongsBelow` are gone: with the model corrected the first collapses to it and the second is unreachable, since it only fires for a newly added absolute path and `selectTargetBlockIndex` returns block 0 for those before the upper bound is read.

`importRank` keeps its three values — `couldResolveAgainst` still reads the bare-vs-dotted distinction — but only rank 0 decides an ordering now.

## Left unchanged, deliberately

`ImportDocumentEditor.couldResolveAgainst`. It is pinned-only and encodes an intentional override: the import being added is read as the provider a pinned dotted import is missing, which is the one case where writing it above an existing import is right. Against the corrected model that is now an explicit override rather than a consequence of the rank order, and its doc says so. Agreed with the maintainer before implementing; filed separately rather than decided here.

The second commit fixes a regression the first introduced against that override: a pinned import raising the ceiling also raised the floor, so `blockIsWithin` refused every block — including one lying wholly above the ceiling, which satisfies both the ceiling and the written order. One pinned import now raises one bound.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
Tests:       813 passed, 813 total
Snapshots:   0 total
Time:        8.207 s, estimated 9 s
Ran all test suites.

$ npm run format:check
> prettier --check "**/*.ts"
Checking formatting...
```

Both new regression tests were checked against the unfixed source. `sortImportsByRank(["GameSystems.Inventory", "Weapons"])` and the block-selection case fail on `origin/main`; the ceiling case fails on this branch's first commit.

## Review

One round, fresh context, `PASS_WITH_SUGGESTIONS` — 0 Critical, 4 Important, 5 Suggestion. The four Important findings are addressed in the second commit, except one left out of scope and noted below: the ceiling/floor regression, a false sentence in the `importRank` doc, a self-contradicting `PinnedBounds` doc, and a test whose fixture no longer exercised the case it was named for.

Two findings are outside this ticket and want issues of their own:

- With `importGrouping` set to `digestFirst` or `localFirst`, the grouping branch picks a target block by digest/local purity and never asks which imports are already written where, so a new relative import can still land above an existing relative one. Behaviour is identical on `main`; the corrected model is what turns it into a wrong answer.
- With `preserveImportLocations` off, the consolidation branch hoists every rewritable import to the top with no pinned-import bound at all. Pre-existing and unchanged here.

### DIFF
--- a/changelog.d/271.fixed.md
+++ b/changelog.d/271.fixed.md
@@ -1,0 +1,1 @@
+**Import sorting**: Sorting and adding imports now keep every relative `using` in its written order, since a dotted import can bring into scope the module a later bare import names.
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -147,18 +147,21 @@ function pinnedSpanEnd(imp: ScannedImport, classifications: LineClassification[]
  * bound placement the same way, for the same reason: the line stays put, so
  * whichever of the two put it there changes nothing about what may cross it.
  *
- * Both bounds read the ranks selectTargetBlockIndex compares, so pinned and
- * blocked imports cannot disagree about what precedes what:
+ * Both bounds ask what selectTargetBlockIndex asks, so pinned and blocked
+ * imports cannot disagree about what precedes what:
  *
- * - `floor` - the last line owned by the lowest pinned import that belongs
- *   above the new one (belongsAbove). The new import goes below it.
+ * - `floor` - the last line owned by the lowest pinned import, where the new
+ *   path resolves against what is in scope above it
+ *   (ImportFormatter.resolvesAgainstScopeAbove). Any of them can be bringing
+ *   its first segment into scope, so it goes below all of them. An absolute
+ *   path needs nothing above it and is unconstrained.
  * - `ceiling` - the first line of the highest pinned import that could be
  *   resolving against the new one (couldResolveAgainst). The new import goes
  *   directly above it.
  *
- * The two predicates are deliberately not mirror images: see
- * couldResolveAgainst for why an upper bound here has to be a scope
- * requirement rather than the ordering preference belongsBelow also captures.
+ * The two are deliberately not mirror images: the floor is what every rule
+ * here keeps, and the ceiling is the one documented override of it. See
+ * couldResolveAgainst, and placementLine for which wins.
  *
  * `-1` means unconstrained in that direction.
  */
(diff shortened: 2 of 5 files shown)


vukefn:
Follow-ups from this PR's review, filed: #313 (the pinned-import ceiling), #314 (block choice under importGrouping), #315 (consolidation hoisting past a pinned line).

PR #319, opened 2026-08-11
fix(imports): count a using: pair opened after a definition as present

### LINKED ISSUE #309: A using: pair opened after a non-using definition never counts as present
A `using:` opening an indented pair after a non-using statement is legal Verse, since `;` separates definitions in a scope exactly as a newline does:

```
X := 1; using:
    /Verse.org/Random
```

`scanModuleImports` records nothing for it. The classification reads a line from its head (`ImportFormatter.isModuleImport`, `ImportFormatter.ts:107`), so the line is rejected, and the reject path records only the complete `using` statements the line writes — a `using:` carries no path of its own, and the path below it is on an indented line the scan skips.

Confirmed on `fix/273-using-after-statement` @ `3a34ea4`, which is #273's fix and does not change this shape:

```
scanModuleImports(["X := 1; using:", "    Economy.Shop", "code()"]) -> []
allUsingPaths(["X := 1; using:", "    Economy.Shop", "code()"])     -> ["Economy.Shop"]
```

`allUsingPaths` sees it, so removal safety is unaffected and the line is never touched — the same shape #273 had. The cost is that a diagnostic asking for that path writes a second copy of an import the file already makes.

This was scoped out of #273 deliberately. Recording it means consuming the indented line as part of the entry and deciding the span, which the branch handling the same pair after a *`using`* statement already does (`ImportScanner.ts`, the `\busing\s*:\s*$` branch) — but that branch is reachable only because the line's head passed the classification. Reaching it from a rejected line needs the pinning and span rules worked out for a line the scanner may not rebuild.

The `scanModuleImports` doc comment states the exclusion, so the current behaviour is documented rather than silent.

## Suggested acceptance

- The path below a `using:` opened after a non-using statement counts as present for deduplication.
- Its line and the indented path line stay exactly as written, as they do today.

vukefn:
Closes #309

## What changed

`scanModuleImports` recorded nothing for a `using:` pair opened after a non-using definition, `X := 1; using:` over an indented path. The classification reads a line from its head, so the line is rejected, and the reject path recorded only the complete `using` statements the line writes — a `using:` carries none of its own, and the path below it is on a line the scan skips. Nothing counted that path as present, so a diagnostic asking for it wrote a second copy of an import the file already makes.

The reject path now records the pair as well, spanning both lines and pinned, so neither line is ever rebuilt or moved. `allUsingPaths` already saw the path, so removal safety was never affected and the line was never at risk.

Two rules decide what the new branch admits:

- **Content** is classified by handing `isModuleImport` the matched `using:` and the line below it, so a pair is admitted on the same content rule the branch for a pair after a `using` applies. Recording it unclassified would have widened the plain pair too — content the classification declines, `using:` over `Foo'Loc'`, reaches the reject path by the same rejection.
- **The path line must read as live code.** The tail is matched against the line's code, so it is found whatever trivia follows the `using:` — which admits `using: # note` and `using: <#>` as well as a definition-headed line. Under `<#>` the path is comment text, and counting it as imported would withhold an import the file genuinely needs, the mirror of the gap being closed. A line comment ends with its line, so `using: # note` over a path does count, pinned because a rebuild drops the opener's comment.

## Acceptance criteria

| Criterion | Where |
|---|---|
| The path below such a pair counts as present for deduplication | `records the path below a using: opened after a definition, pinned`; `counts the path below a using: opened after a definition as already present` |
| Both lines stay exactly as written | `offers no rewritable import for a pair opened after a definition` |

## Verse semantics

The shape is legal Verse, confirmed against the compiler grammar rather than inferred: `Block := Brace | DotSpace Space Def Space | (DotSpace | ':') Space Ind List Ded`, and `Ind := Ending push; set Nest=false; set BlockInd=LineInd` sets the block body's indent reference to the whole line's leading whitespace, whatever text preceded the `:`. With `;` separating scope elements, `X := 1; using:` over an indented path parses as two elements of the file-scope list.

## Not in scope

The branch for a pair opened after a `using` statement answers two of these questions differently: it counts the comment text below a `using { /A }; using: <#>` as a path, and it applies no content rule to the pair, so it records a `Foo'Loc'` this branch declines. Both divergences reproduce identically on `main`, byte for byte, and are untouched here. They belong in their own issue rather than in this diff.

## Review

Two rounds, fresh context each, against the smell baseline and both acceptance criteria.

- **Round 1** — PASS_WITH_SUGGESTIONS, 1 Important, 3 Suggestions. The Important was a real regression this diff introduced: matching the `using:` tail against the line's code admits openers the head check rejected for trailing trivia alone, so `using: <#>` counted its commented-out path as imported. Fixed by requiring the path line not to continue a comment opened above it. Two suggestions taken as tests; one declined as scope widening.
- **Round 2** — PASS_WITH_SUGGESTIONS, 1 Important, 5 Suggestions. The Important was comment accuracy, not behaviour: the branch comment said "the path line has to read as live code", which is a wider predicate than the guard applies, and implementing it literally would have admitted `using: <#` over `note #> /A` with the suite still green. Reworded to name the actual check, with the two boundary cases pinned by tests. The remaining suggestions were the out-of-scope divergences above and one informational note on an unterminated-literal shape that `LineScan.codeOutsideLiterals` already documents as accepted.

## Verification

`npm run compile`:

```
> tsc -p ./
```

`npm test`:

```
> jest --verbose

Tests:       831 passed, 831 total
Snapshots:   0 total
Time:        8.585 s
Ran all test suites.
```

`npm run format:check`:

```
> prettier --check "**/*.ts"

Checking formatting...
```

Each new regression test was proved to detect the defect: with the fix removed, the four tests asserting the new behaviour fail and the guard tests still pass.

### DIFF
--- a/changelog.d/309.fixed.md
+++ b/changelog.d/309.fixed.md
@@ -1,0 +1,1 @@
+**Import scanning**: An indented `using:` pair opened after another statement on the same line now counts as already imported, so a suggested import is no longer added a second time.
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -386,8 +386,11 @@ export function classifyLines(lines: string[]): LineClassification[] {
  *   complete `using` statement it writes is recorded, pinned, because the path
  *   is imported and writing it again would duplicate it. The pair a `using:`
  *   opens at the end of such a line is recorded with them, spanning both
- *   lines, on the same ground - the path below it is imported, and the
- *   classification would have admitted it had the `using:` headed the line.
+ *   lines, on the same ground - the path below it is imported. That tail is
+ *   read from the line's code, so a `using:` the classification refused only
+ *   for the trivia after it opens a pair here as well; the path line must
+ *   itself read as live code, or a `using: <#>` would count the comment text
+ *   below it as an import.
  * - A path is only ever read from the line's statements, never from the text of
  *   a `"` string it writes. A `'` is not treated as opening literal text here,
  *   because it also closes a path's quoted segment suffix; see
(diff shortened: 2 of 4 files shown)


vukefn:
Follow-up filed as #320: the branch for a pair opened after a `using` statement counts a commented-out path as present, and applies no content rule to the pair. Both reproduce byte-for-byte on `main` and are untouched here.

PR #323, opened 2026-08-11
fix(services): drop a skipped module's subtree with its parent

### LINKED ISSUE #316: A skipped module's descendants are promoted to file scope instead of dropped with it
Found while fixing #305, and pre-existing to it. Filed apart because it predates that ticket and is reachable today without it.

## What happens

`ProjectPathScanner.extractDeclarations` skips a module it judges unimportable, but `continue`s before the module reaches `moduleStack`:

`src/services/ProjectPathScanner.ts:220-226`:

```ts
if (shouldSkipDeclaration(visibility, true)) {
    continue;
}

const fullPath = currentModulePath ? `${currentModulePath}.${name}` : name;
moduleStack.push({ name: fullPath, indent });
```

The stack is what supplies the enclosing segment of every `fullPath` below, so what the skipped module contains does not go away with it - it is promoted to file scope:

| Parent specifier | Nodes recorded |
| --- | --- |
| `<public>` | `Systems`, `Systems.Inner`, `Systems.Inner.X` |
| `<private>` | `Inner`, `Inner.X` |
| `<scoped{ModuleA}>` (after #305) | `Inner`, `Inner.X` |

`<private>` reads that way today; #305 routes `scoped` and `epic_internal` into the same path, which is why the behaviour is now pinned by `promotes a skipped module's descendants to file scope` in `ProjectPathScanner.test.ts` rather than left undocumented.

## Why it matters

Importing a module checks the accessibility of every path segment from root to target, so a subtree under an unimportable parent is exactly as unreachable as the parent. Dropping the parent and keeping its children inverts the intent stated at `ProjectPathScanner.ts:153-156` - "Anything narrower cannot be imported from outside."

Two consequences, both worse than keeping the whole subtree:

- The recorded path does not exist. `resolveModuleLocations("Systems/Inner")` answers with nothing, while `resolveModuleLocations("Inner")` answers with a location the compiler will reject.
- A genuine top-level `Inner` elsewhere in the project gains a spurious second candidate at the same name, which is the ambiguous "which of these did you mean" prompt rather than a silent miss.

## Fix

Let a skipped module take its subtree with it. It has to reach `moduleStack` either way, because the stack is also what tracks indentation:

```ts
const moduleStack: { name: string; indent: number; skipped: boolean }[] = [];
```

Push it marked skipped, then skip every declaration with a skipped module open above it, after the pop loop at `ProjectPathScanner.ts:209-211` and before any pattern match. Grandchildren need no separate handling: they are deeper than the skipped entry, so it is still open when they are read.

`includePrivate` must keep lifting this along with the two checks it already lifts, or a caller asking for everything would lose the subtree instead of gaining it.

## Coverage

`promotes a skipped module's descendants to file scope` pins today's behaviour for `<scoped{ModuleA}>` and `<private>`; it inverts when this is fixed. Worth adding: a grandchild two levels down, a sibling declaration after the skipped module closes (which must still be recorded), and the `includePrivate: true` path.

vukefn:
Closes #316

`extractDeclarations` skipped a module it judged unimportable but `continue`d before the module reached `moduleStack`. The stack supplies the enclosing segment of every `fullPath` below it, so the subtree did not go away with its parent — it was promoted to file scope, producing a path the compiler rejects and a spurious second candidate for any genuine top-level module of the same name.

A skipped module now reaches the stack marked `skipped`, holding its indent, and every declaration under one is dropped with it. Grandchildren need no separate handling: they are deeper than the skipped entry, so it is still open when they are read. `includePrivate` lifts this along with the two checks it already lifted, so a caller asking for everything gains the subtree nested rather than losing it.

## Behaviour change

For `Systems<private or scoped{...}> := module:` containing `Inner<public> := module:` containing `X`:

| | Before | After |
| --- | --- | --- |
| Recorded | `Inner`, `Inner.X` | nothing |
| `includePrivate: true` | `Systems`, `Systems.Inner`, `Systems.Inner.X` | unchanged |

## Tests

`promotes a skipped module's descendants to file scope` inverted to `drops a skipped module's descendants along with it`. Added: a grandchild two levels down, a sibling declared after the skipped module closes, the `includePrivate: true` path, and a skipped module nested inside a kept one — the case where the pop unwinds back to a kept parent, so a wrong pop bound surfaces as a missing prefix rather than a dropped node.

The first three fail against the unfixed scanner, each returning the promoted `Inner` / `Inner.Deep` paths. The `includePrivate` case passes before and after by design: it guards the lift rather than detecting the defect.

## Verification

```
> tsc -p ./
```

```
> jest --verbose

Tests:       849 passed, 849 total
Snapshots:   0 total
Time:        12.629 s
Ran all test suites.
```

```
> prettier --check "**/*.ts"

Checking formatting...
```

## Review

One round, fresh context, verdict `PASS_WITH_SUGGESTIONS` — no Critical findings. The reviewer compiled the scanner and ran inputs through it rather than reading alone, confirming: the "nothing is pushed above a skipped entry" invariant holds, so the top-of-stack test is equivalent to scanning the stack; no stale `currentModulePath` leaks past the gate; a `#` comment at column 0 inside a skipped module does not close it; tab-indented bodies drop; consecutive skipped siblings do not swallow the following kept module.

Two findings noted and deliberately not taken here, both pre-existing and outside this ticket:

- **The next-line brace form still promotes its subtree.** `Systems<private> := module` with `{` on the following line leaks `Inner` and `Deep` to file scope. The bare `{` sits at the declaration's own indent, so the existing pop rule closes the skipped entry before the body is read. That rule is already pinned for a public parent by `leaves the members of a next-line brace module unnested`, and the prescribed mechanism in #316 is implemented to the letter. Worth its own ticket against the pop rule.
- **`currentModulePath = fullPath` is a dead store** — line 229 recomputes it from the stack top before any read. Pre-existing; removing it is a refactor this ticket did not ask for.

### DIFF
--- a/changelog.d/316.fixed.md
+++ b/changelog.d/316.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: The declarations nested inside a module that cannot be imported are no longer offered as import candidates under a shortened path that does not compile.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -133,7 +133,11 @@ export class ProjectPathScanner {
         const lines = maskCommentsAndStrings(content).split("\n");
 
         let currentModulePath = "";
-        const moduleStack: { name: string; indent: number }[] = [];
+        // A skipped module is on the stack for its indent alone, so that what
+        // it contains is dropped with it rather than losing the enclosing
+        // segment from its path. Nothing is ever pushed above a skipped entry,
+        // because every declaration under one is skipped before it can be.
+        const moduleStack: { name: string; indent: number; skipped: boolean }[] = [];
 
         // A declaration carries any number of stacked specifiers, of which at
         // most one is a visibility; the rest (<native>, <final>, ...) are noise
@@ -213,6 +217,15 @@ export class ProjectPathScanner {
             while (moduleStack.length > 0 && indent <= moduleStack[moduleStack.length - 1].indent) {
                 moduleStack.pop();
             }
+
+            // An import checks every path segment from the root, so anything
+            // still inside a skipped module is as unreachable as the module
+            // and is dropped with it. The pop loop above has already closed
+            // the module for a line that left it.
+            if (moduleStack.length > 0 && moduleStack[moduleStack.length - 1].skipped) {
+                continue;
+            }
+
             currentModulePath = moduleStack.length > 0 ? moduleStack[moduleStack.length - 1].name : "";
 
             const moduleMatch = line.match(modulePattern);
@@ -221,12 +234,14 @@ export class ProjectPathScanner {
                 const visibility = extractVisibility(specifiers);
                 const isPublic = visibility === "public";
 
+                const fullPath = currentModulePath ? `${currentModulePath}.${name}` : name;
+
                 if (shouldSkipDeclaration(visibility, true)) {
+                    moduleStack.push({ name: fullPath, indent, skipped: true });
                     continue;
                 }
 
-                const fullPath = currentModulePath ? `${currentModulePath}.${name}` : name;
-                moduleStack.push({ name: fullPath, indent });
+                moduleStack.push({ name: fullPath, indent, skipped: false });
                 currentModulePath = fullPath;
 
                 nodes.push({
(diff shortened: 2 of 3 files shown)


vukefn:
The next-line brace gap noted in the description is now tracked as #324. It is pre-existing and belongs to the pop rule rather than to the skip path this PR changed, so it stays out of this diff.

PR #329, opened 2026-08-11
fix(services): keep a module open across its own next-line brace

### LINKED ISSUE #324: A next-line brace module is closed by its own opening brace, promoting its members
Found by the review gate on #323, and pre-existing to it. Filed apart because the fix belongs to the pop rule rather than to the skip path #316 changed.

## What happens

`ProjectPathScanner.extractDeclarations` closes an open module on indentation alone:

`src/services/ProjectPathScanner.ts:215-219`:

```ts
// Indentation alone closes a module: a line at or left of the open
// module's own indent is outside it.
while (moduleStack.length > 0 && indent <= moduleStack[moduleStack.length - 1].indent) {
    moduleStack.pop();
}
```

A module written in the next-line brace form puts its opening `{` at the declaration's own indent, so that brace pops the module before any of its body is read:

```verse
Systems<private> := module
{
    Inner<public> := module{}
    Deep<public> := class {}
}
```

Verified against the compiled scanner, this records `Inner` at file scope and `Deep` after it — the module's own indent-0 brace closed it on line 2.

## Why it matters

For a `<public>` parent this is a known limit, pinned by `leaves the members of a next-line brace module unnested` in `ProjectPathScanner.test.ts`: the members lose their enclosing segment but stay reachable, so the recorded paths are merely shorter than they should be.

For a parent the scan judges unimportable it is the failure #316 fixed, reachable by another route. #316 made a skipped module hold its subtree by keeping it on the stack, but that only works while the module is still open, and here it is closed a line early. So the subtree is promoted to file scope again: a path the compiler rejects, plus a spurious second candidate for any genuine top-level module of the same name.

The same-line brace form (`module {`), the dotted form (`module.`) and the colon form all drop the subtree correctly. Only the next-line brace leaks.

## Fix

The pop rule needs to not treat a module's own opening brace as a line that left the module. Two shapes worth weighing:

- Defer the pop by one line when the previous line was a module declaration whose body had not yet opened, so the brace is read as belonging to that declaration.
- Track the brace depth alongside the indent, so a braced module closes on its `}` rather than on indentation.

The second is the larger change and would also fix the `<public>` case the existing test pins, so that test inverts with it. Either way `leaves the members of a next-line brace module unnested` states today's behaviour and has to be revisited rather than worked around.

## Coverage

A next-line brace module under each of `<private>`, `<scoped{...}>` and `<public>`, with a grandchild, and a sibling declared after the closing `}`.

vukefn:
Closes #324

## What was wrong

`ProjectPathScanner.extractDeclarations` closed a module on indentation alone. A module written in the next-line brace form puts its opening `{` at the declaration's own indent, so that brace read as a line that had left the module and popped it before any member was scanned.

For a kept parent the members lost their enclosing segment, so recorded paths were merely shorter than they should be — the limit the test `leaves the members of a next-line brace module unnested` pinned. For a parent the scan judges unimportable it was the failure #316 fixed, reached by another route: #316 holds a skipped module's subtree by keeping the module on the stack, which only works while the module is still open, and here it was closed a line early. The subtree was promoted to file scope again, offering paths the compiler rejects plus a spurious second candidate for any genuine top-level module of the same name.

## The change

The ticket weighed two shapes. This takes the first, the deferred pop, rather than full brace-depth tracking: a module declaration whose line ends without opening a body now waits one considered line for its brace, and that brace no longer reaches the pop rule. Blank lines and comments never reach the pop rule either, so the wait tolerates the ones Verse permits between the declaration and its brace.

Both fix shapes invert the pinned test, as the ticket noted, so it is renamed and its expectation corrected rather than worked around.

## One correction to the ticket's coverage list

The ticket asked for a next-line brace module under each of `<private>`, `<scoped{...}>` and `<public>`. `<private>` on a module is not legal Verse — `assert_semantic_error(3594){ M<private> := module {} }` in the Book of Verse `Tests/Modules/Module.versetest:36`, "Access levels protected and private are only allowed inside classes." A module can never sit inside a class, so that is unconditional. `<scoped{...}>` on a module is legal, so it carries the unimportable-parent case and the `<private>` case is dropped.

The next-line brace form itself is confirmed legal: the Book of Verse carries it as a compile-validated versetest at `docs/16_modules.md:128-131`, labelled "Bracket syntax (also supported)".

## Coverage

Six tests, five of which fail against the unfixed scanner:

- `nests the members of a next-line brace module` — the inverted pin
- `closes a next-line brace module at its closing brace` — a sibling after `}` is not swallowed
- `still waits for the brace across the blank lines and comments Verse allows before it`
- `drops the subtree of a skipped module whose body opens on the next line` — `<scoped{...}>` parent, grandchild, sibling after `}`
- `nests a kept module's subtree through a next-line brace body` — `<public>` parent, grandchild, sibling
- `closes a module whose declaration line opened no body and no brace followed` — passes both ways by design; it guards against the wait over-holding a module, so it is a guard rather than a regression test for this defect

Each of the five was confirmed to fail with the fix reverted and the tests left in place.

## Known limitation, left alone

A braced body written at or left of the module's own indent still pops early. Verse permits it — the Book of Verse's own example has `m := module{` at indent 0 with its body at indent 0 — but this is pre-existing and shared with the same-line brace form, so it is outside this ticket. Fixing it is the brace-depth tracking the ticket weighed as the larger second shape.

## Review

One round, verdict `PASS_WITH_SUGGESTIONS` — no Critical, no Important. The reviewer reproduced the build, the suite and the revert-and-rerun, and probed the shapes the fix could plausibly break: a trailing `#` comment on the declaration line, a `<# ... #>` block spanning the gap, the `<#>` indented marker, tabs, a stray `using` between declaration and brace, nested next-line braces, and a module as the last line of a file. All nest correctly.

It also settled a property the change rests on but did not state: a line that triggers the wait can never itself record a node, because every declaration pattern is anchored `^(\w+)` or `^\([^)]+\)\.` and none can match a leading `{`. So suppressing the pop loop on that line cannot mis-attribute anything, and only the top of stack can ever be awaiting a brace.

Both suggestions were addressed: the pop-loop snapshot is renamed and commented for when it is read (second commit), and the residual limitation is documented above rather than fixed here.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Tests:       866 passed, 866 total
Snapshots:   0 total
Time:        42.084 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/324.fixed.md
+++ b/changelog.d/324.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: A module whose body opens with a brace on the next line now nests its members, so their import paths carry the enclosing module and an unimportable one no longer offers its contents as top-level candidates.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -137,7 +137,11 @@ export class ProjectPathScanner {
         // it contains is dropped with it rather than losing the enclosing
         // segment from its path. Nothing is ever pushed above a skipped entry,
         // because every declaration under one is skipped before it can be.
-        const moduleStack: { name: string; indent: number; skipped: boolean }[] = [];
+        //
+        // `awaitingBrace` is true only between a module declaration that ended
+        // without opening a body and the next line considered, which is where
+        // its `{` may be.
+        const moduleStack: { name: string; indent: number; skipped: boolean; awaitingBrace: boolean }[] = [];
 
         // A declaration carries any number of stacked specifiers, of which at
         // most one is a visibility; the rest (<native>, <final>, ...) are noise
@@ -184,8 +188,10 @@ export class ProjectPathScanner {
         // how a next-line brace reaches a scan that sees one line at a time.
         // `>` holds parity with the two other spellings of this grammar,
         // ImportPathConverter.buildModuleDefinitionRegex and
-        // moduleDeclarations' MODULE_DECLARATION.
-        const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*(?:[:>{.]|$)/;
+        // moduleDeclarations' MODULE_DECLARATION. [3] captures which of them
+        // ended the keyword, and is undefined for the line end alone; the
+        // module stack needs that to know a body is still to open.
+        const modulePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*module\s*(?:([:>{.])|$)/;
         const classPattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*class\s*(?:<[^>]+>)*\s*[\(:]?/;
         const structPattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*struct\s*(?:<[^>]+>)*\s*[\(:]?/;
         const interfacePattern = /^(\w+)((?:<[^>]+>)*)\s*:=\s*interface\s*(?:<[^>]+>)*\s*[\(:]?/;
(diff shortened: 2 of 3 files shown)


vukefn:
The residual limitation noted under "Known limitation, left alone" is now tracked as #330. It is pre-existing and shared with the same-line brace form, so this PR neither introduces nor widens it; closing it needs the brace-depth tracking #324 deferred.

PR #332, opened 2026-08-12
fix(services): close a braced module body on its brace, not on indentation

### LINKED ISSUE #330: A braced module body written at or left of its own indent promotes its members
Found by the review gate on #329, and pre-existing to it. Filed apart because closing it needs the brace-depth tracking #324 weighed and deferred.

## What happens

`ProjectPathScanner.extractDeclarations` closes an open module on indentation alone:

`src/services/ProjectPathScanner.ts`:

```ts
// Indentation alone closes a module: a line at or left of the open
// module's own indent is outside it.
while (moduleStack.length > 0 && indent <= moduleStack[moduleStack.length - 1].indent) {
    moduleStack.pop();
}
```

Inside a braced body Verse does not require the members to be indented past the declaration. When they are not, the first member pops the module that encloses it:

```verse
Systems<scoped{ModuleA}> := module
{
B<public> := module:
    X<public> := class {}
}
```

Both this and the same-line form (`Systems<scoped{ModuleA}> := module{`) record `B` and `B.X` at file scope. The outcomes are identical, which is what makes this a separate defect from #324 rather than a leftover of it: #324 was the module's own opening brace closing it, and #329 fixed that; this is the body's own members closing it, and it reaches the same failure by a different line.

The shape is legal Verse. The Book of Verse's own module example writes it — `docs/16_modules.md:120-134`, a compile-validated versetest, opens `m := module{` at indent 0 and writes `module1 := module:` at indent 0 inside it.

## Why it matters

For a `<public>` parent the members lose their enclosing segment, so the recorded paths are shorter than they should be but stay reachable.

For a parent the scan judges unimportable it is the #316 failure class: the subtree is promoted to file scope, offering paths the compiler rejects plus a spurious second candidate for any genuine top-level module of the same name. #316 holds a skipped module's subtree by keeping the module on the stack, and that only works while the module is still open.

## Fix

Indentation cannot close a braced body, so the module stack needs to know how each entry's body was opened and close it accordingly: a braced module on its matching `}`, a colon module on indentation as now. This is the brace-depth tracking weighed as the second shape in #324 and deferred there as the larger change.

The counter has to be brace depth relative to where the module's body opened, not a global count, because class, struct, function and map bodies all contribute braces of their own. Comments and string literals are already masked before the scan, so their braces do not reach it.

## Coverage

A braced module body whose members sit at or left of the declaration's own indent, in both the same-line and next-line brace forms, under `<public>` and under `<scoped{...}>`, with a grandchild and a sibling declared after the closing `}`. The mixed case is worth pinning too: a colon-form module containing a braced one, and the reverse, so the two closing rules are shown not to interfere.

vukefn:
Closes #330

The module stack closed every entry on indentation. Verse delimits a braced body by its braces alone, and lets its members sit at or left of the declaration's own indent — the Book of Verse writes it that way in a compile-validated example (`docs/16_modules.md:120-132`, `m := module{` holding `module1 := module:`, both at indent 0). The first such member popped the module enclosing it, so the whole body was promoted to file scope: paths one segment short under a kept parent, and under a skipped one a subtree the compiler rejects, offered beside any genuine top-level module of the same name.

#324 weighed brace-depth tracking and deferred it; #329 fixed the module's own opening brace closing it. This is the remaining half — the body's members closing it.

## The change

A `moduleStack` entry now carries `closeDepth`: the brace depth just outside a braced body, or `null` where indentation is what closes the entry (the colon, dotted and `>` forms, and a braced one whose `{` has not been seen yet). One pop loop applies whichever rule the entry was opened with, and stops at the first entry still open — which is what keeps indentation from closing an outer module while a braced one nested inside it is still waiting for its `}`.

Brace depth is counted on the masked lines, so comments and string literals contribute none, and it is counted ahead of the `using` skip: a braced `using` clause spans lines, and skipping its opener while still counting its `}` would close the module around it one brace early.

The grammar behind this is `Block := Brace | DotSpace Space Def Space | (DotSpace | ':') Space Ind List Ded` (`VerseGrammar.h:2346`) — the `Ind`/`Ded` pair appears only in the colon/dot alternative, so no indentation rule reaches a braced body.

## Char literals

`maskCommentsAndStrings` deliberately does not track the single quote, because a quote opens a char literal and an identifier's quoted suffix alike. Left alone, `X:char = '{'` would leak a brace into the depth counter — and unlike the old indentation rule, which recovered at the next dedent, a stranded depth never recovers, so one character would misplace every later declaration in the file. That is a regression this change would otherwise introduce, not carried-over debt, so it is fixed here rather than accepted: a brace between single quotes is skipped when counting. A quoted suffix cannot contain a brace (`Ident := ... "'" {!('<#'|'#>'|'\'|'{'|'}'|'"'|''') ...} "'"`, `VerseGrammar.h`), so a brace between quotes is unambiguously a literal and never a suffix.

One narrower hole is left, and it belongs to the masker rather than to this counter: a char literal holding a double quote (`Print('"')`) opens a string in the mask and blanks the rest of that line, taking any real brace on it. Making the masker track quotes means telling a char literal from a quoted suffix, which is the ambiguity it documents as the reason it does not.

## Tests

Eight new cases under `braced module bodies`, covering exactly what the ticket asked: members at and left of the declaration's own indent, in both the same-line and next-line brace forms, under `<public>` and under `<scoped{...}>`, with a grandchild and a sibling after the closing `}`; the mixed case in both directions (a colon module holding a braced one and the reverse) to show the two closing rules do not interfere; and a multi-line braced `using` clause inside a braced module.

Each was run against the unfixed scanner first — all seven of the original cases fail there, with the reported symptom (`Systems.B` recorded as `B`); the char-literal case fails with the guard removed (`Loose` recorded as `Systems.Loose`).

## Review

Reviewed in a fresh context at the `models.review` tier, over roughly sixty probes run against both this branch and its baseline. Verdict `PASS_WITH_SUGGESTIONS`, no `Critical`. The one `Important` was the char-literal regression above, which is fixed rather than accepted; two comments that named constraints the code does not actually impose were corrected, and the same-line `<public>` case gained the trailing sibling the ticket asks for, so it now asserts the module closes.

One finding is deliberately left alone as out of scope: `src/visibility/moduleDeclarations.ts` runs a second, independent scan that closes module nesting on indentation with the same rule this PR replaces, and returns `[["Systems"], ["B"]]` for the ticket's own snippet. This PR does not make it worse, and the ticket names only `ProjectPathScanner`. Worth its own issue, since a visibility edit driven by the wrong chain writes into the wrong module.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Tests:       900 passed, 900 total
Snapshots:   0 total
Time:        9.233 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/330.fixed.md
+++ b/changelog.d/330.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: A module whose braced body sits at or left of its own indent now nests its members, so their import paths keep the enclosing module and an unimportable one no longer offers its contents as top-level candidates.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -115,8 +115,9 @@ export class ProjectPathScanner {
      * `sourceLine` is 1-based, so it is one more than the `vscode.Position` line
      * for the same declaration. `fullPath` carries the dotted chain of enclosing
      * modules, so it equals `name` only at file scope. Declarations nested in a
-     * class or struct body are not filtered out here; only indentation-scoped
-     * module nesting is tracked.
+     * class or struct body are not filtered out here; only module nesting is
+     * tracked, each module scoped by its braces or by its indentation according
+     * to how its body was opened.
      *
      * Comments and string literals are masked before the scan, so a
      * commented-out declaration is neither recorded nor pushed onto the module
@@ -133,6 +134,11 @@ export class ProjectPathScanner {
         const lines = maskCommentsAndStrings(content).split("\n");
 
         let currentModulePath = "";
+        // Brace depth across the file, counted on the masked lines so comments
+        // and string literals contribute none. Verse delimits a braced body by
+        // its braces alone, so this is the only thing that can close one: its
+        // members are free to sit at or left of the declaration's own indent.
+        let braceDepth = 0;
         // A skipped module is on the stack for its indent alone, so that what
         // it contains is dropped with it rather than losing the enclosing
         // segment from its path. Nothing is ever pushed above a skipped entry,
@@ -140,8 +146,10 @@ export class ProjectPathScanner {
         //
         // `awaitingBrace` is true only between a module declaration that ended
         // without opening a body and the next line considered, which is where
-        // its `{` may be.
-        const moduleStack: { name: string; indent: number; skipped: boolean; awaitingBrace: boolean }[] = [];
+        // its `{` may be. `closeDepth` is the brace depth just outside a braced
+        // body, and null while indentation is what closes the entry: the colon,
+        // dotted and `>` forms, and a braced one whose `{` is still to come.
+        const moduleStack: { name: string; indent: number; skipped: boolean; awaitingBrace: boolean; closeDepth: number | null }[] = [];
 
         // A declaration carries any number of stacked specifiers, of which at
         // most one is a visibility; the rest (<native>, <final>, ...) are noise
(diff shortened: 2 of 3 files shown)


vukefn:
The same defect in the second scan is filed as #333: `findExplicitModuleDeclarations` still closes a module on indentation, so it returns `[["Systems"], ["B"]]` for this ticket's own snippet. Out of scope here, and this PR does not make it worse, but the two scans now disagree about the same file until it lands.

PR #341, opened 2026-08-12
fix(services): leave a closed braced body the modules inside it shield

### LINKED ISSUE #336: The project scan closes a braced module body only from the top, so a module inside it can shield the closed brace
Found by the review gate on #333, after PR #335 landed the fix in the other scan.

`ProjectPathScanner` and `findExplicitModuleDeclarations` both read module nesting out of the same file, and they are meant to agree. #332 fixed the scanner to close a braced body on its brace; #335 fixed the same defect in the visibility scan, and went one step further. The scanner did not take that step, so the divergence has flipped: the visibility scan is now right and the scanner is the only one carrying the bug.

## What happens

Closing only from the top of the module stack lets an indentation-closed entry shield a braced one beneath it. A colon module written at a *lower* indent than the braced declaration it sits inside — the shape #330 and #333 exist to support — reads as still open at any declaration indented past it, so the pop loop stops there and never reaches the braced entry whose `}` has already returned the depth to where its body opened.

`src/services/ProjectPathScanner.ts:266-292` pops from the top only:

```ts
while (moduleStack.length > 0) {
    const openModule = moduleStack[moduleStack.length - 1];
    const closed = openModule.closeDepth !== null ? depthAtLineStart <= openModule.closeDepth : indent <= openModule.indent;
    if (!closed) {
        break;
    }
    moduleStack.pop();
}
```

The scanner escapes the simpler shapes because it also runs this loop on the `}` line itself, which the offset-driven visibility scan has no equivalent of. That does not generalise: it only closes the innermost entry, so a shielding entry above a closed brace still survives.

## The six diverging shapes

Each is `content` → what `ProjectPathScanner` reports, versus what `findExplicitModuleDeclarations` now reports. The second is correct in every case.

| Input | Scanner | Visibility scan |
| --- | --- | --- |
| `"Outer := module:\n    A := module{\nB := module:\n    Q:int = 1\n    }\n    C := module {}\n"` | `Outer.A.B.C` | `Outer.C` |
| `"L1 := module:\n    L2 := module:\n        A := module{\nB := module:\n    Q:int = 1\n        }\n        C := module {}\n"` | `L1.L2.A.B.C` | `L1.L2.C` |
| `"Outer := module:\n    A := module{\nB := module:\n    Q:int = 1\n    }\n        C := module {}\n"` | `Outer.A.B.C` | `Outer.C` |
| `"Outer := module:\n    Inner := module\n    {\nMember := module:\n    Q:int = 1\n    }\n    Kept := module {}\n"` | `Outer.Inner.Member.Kept` | `Outer.Kept` |
| `"Systems := module{\nB := module:\n    C := module{\nD := module:\n        Z:int = 1\n    }\n    E := module {}\n}\nSibling := module {}\n"` | `Systems.B.C.D.E` | `Systems.B.E` |
| `"A := module{\nB := module:\n    C := module{\nD := module:\n    Q:int = 1\n}\n    E := module {}\n}\nZ := module {}\n"` | `A.B.C.D.E` | `A.B.E` |

## Why it matters

A `fullPath` one or more segments too long is an import path for a module that does not exist at that path, so the import the extension offers does not resolve. Worse under a skipped parent: the stack keeps a skipped module on it so that everything inside is dropped with it, and a subtree wrongly nested under a *different* entry escapes that drop and gets offered as a top-level candidate — the harm #330 was filed for, reachable again by a different route.

## Fix

Port the two-phase close from `src/visibility/moduleDeclarations.ts`:

- `outermostClosedBrace(stack, braceDepth)` — the lowest index whose `closeDepth !== null && braceDepth <= closeDepth`, truncating the stack there, because a closed braced body takes every module inside it along. The lowest index is the cut because braced entries' close depths increase up the stack, which holds only while the truncation runs before the next push.
- `closedByIndent(entry, indent, line)` — false for a braced entry, so the top-down loop stops at a body still awaiting its `}`. Only a braced entry can be closed from below: its depth test holds anywhere in the file, whereas an indent compared across a brace boundary means nothing.

The scanner keeps `currentModulePath` and the `skipped` flag alongside the stack, so the truncation has to restore `currentModulePath` from the entry now on top rather than assume single pops.

**Extract `countBraces` into `src/utils/verseText.ts` in the same change.** The two scans carry line-for-line copies of the brace counter, single-quote carve-out and comment prose, and they have now drifted in both directions in the space of two PRs. `verseText.ts` already hosts the `maskCommentsAndStrings` and `indentOf` both consume, so it is the obvious home.

## Coverage

The six shapes above, asserted on `fullPath`. Plus the hazards #335's review probed and found sound there, which the port must not reintroduce: negative brace depth meeting `braceDepth <= closeDepth` at index 0 (a stray `}` ahead of any module gives a negative `closeDepth`), a stray `}` closing a braced module before a sibling braced one is declared, and truncation discarding an entry it should keep — one colon entry below the outer brace, two colon levels below it, and a sibling more indented than the enclosing body after a close.

vukefn:
Closes #336

The project scan popped its module stack from the top only, so an entry indentation still held open shielded a braced entry beneath it whose `}` had already returned the depth to where its body opened. Everything after it kept that module as a path prefix, and a `fullPath` one or more segments too long is an import path for a module that does not exist there.

## What changed

- `src/services/ProjectPathScanner.ts` — the close runs in two phases, as `findExplicitModuleDeclarations` has since #335. `outermostClosedBrace` cuts the stack at the lowest entry whose braced body has closed, because one takes every module inside it along; what is left pops innermost out through `closedByIndent`, which is false for a braced entry and so stops at a body still awaiting its `}`. The inline stack-entry type is now a named `OpenModule`.
- `src/utils/verseText.ts` — `countBraces` moves here, beside the `maskCommentsAndStrings` and `indentOf` both scans already consume.
- `src/visibility/moduleDeclarations.ts` — drops its own copy of `countBraces` and imports the shared one.
- Tests — the ticket's six shapes asserted on `fullPath`, plus the three retention hazards and the two stray-brace hazards the port had to avoid reintroducing.

`currentModulePath` needed no change: it is already reassigned from whatever entry the stack leaves on top, so a truncation removing several entries at once restores it correctly.

## One correction to the ticket

The ticket's table says row 6 (`A := module{...` → `A.B.E`) is a diverging shape. It is not — the scanner already returns the correct `A.B.E` on `main`. That row escapes because the scanner also runs its close loop on the `}` line itself, which pops the shielding `D` where `D` sits at indent 0 and the `}` is at indent 0. The other five rows reproduce exactly as written. The test for it is kept as a pin rather than a regression.

## Verification

```
npm run compile
> tsc -p ./
(clean)

npm test
Tests:       970 passed, 970 total

npm run format:check
```

The six shapes and both stray-brace hazards were run against the unfixed code first: five of the six failed with exactly the wrong values the ticket predicts.

## Review

One round, verdict PASS_WITH_SUGGESTIONS, no Critical. The reviewer fuzzed 20,000 generated files across all three body styles and free member indentation: 0 wrong after the fix, 558 wrong before it, and 0 disagreements between the two scans — which is what #336 asked for. Two suggestions are applied here: the missing "two colon levels" retention test, and a close-depth invariant that was stated more absolutely than the code guarantees.

One `Important` finding is deliberately not applied. `outermostClosedBrace` is byte-identical to its twin in `moduleDeclarations.ts`, and the reviewer proposed hoisting it into `verseText.ts` beside `countBraces`. It is left local: `verseText.ts` is for reading Verse source as text, and this reads a module stack, so that module is the wrong home for it. Unifying the two scans' nesting logic is #45's job.

## Found while reviewing, not fixed here

`"Priv<private> := module:\n    Inner := module{\nX := module {}\n    }\nY := module {}\n"` returns `X` and `Y` as top-level candidates, where the true path is `Priv.Inner.X`. Nothing is pushed onto the stack inside a skipped module, so the braced `Inner` body is untracked and the skipped colon entry closes on indentation at `X` — a private subtree offered as an import candidate, the harm class #330 was filed for, reached by a different route. Present on `main` before this change and unaffected by it.

### DIFF
--- a/changelog.d/336.fixed.md
+++ b/changelog.d/336.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: A module written inside a braced module body no longer holds that body open past its closing brace, so imports are no longer offered at paths one or more segments too long.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -534,10 +534,11 @@ export class ProjectPathScanner {
  * the declaration that opened them.
  *
  * The lowest such index is the cut rather than merely one of them, because a
- * braced entry records the depth it opens from and every braced entry left on
- * the stack sits below the current depth: their close depths increase upwards,
- * so nothing below the index found here has closed too. Cutting the stack
- * before the next entry is pushed is what keeps that true.
+ * braced entry records the depth it opens from: close depths never decrease
+ * going up the stack, so nothing below the index found here has closed too.
+ * Two entries may record the same depth - the line opening an awaited body
+ * assigns one after this has already run - which the lowest index still
+ * answers, both of them being closed together.
  */
 function outermostClosedBrace(open: readonly OpenModule[], braceDepth: number): number {
     return open.findIndex((entry) => entry.closeDepth !== null && braceDepth <= entry.closeDepth);
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -1,11 +1,38 @@
 import * as vscode from "vscode";
 import { logger } from "../utils";
-import { indentOf, maskCommentsAndStrings } from "../utils/verseText";
+import { countBraces, indentOf, maskCommentsAndStrings } from "../utils/verseText";
 import { ProjectPathHandler } from "../project";
 import { ProjectPathData, ProjectPathNode, ProjectScanOptions } from "../types";
 
 /** Number of .verse files parsed concurrently during a full scan. */
 const SCAN_CONCURRENCY = 8;
+
+/**
+ * A module whose body the scan is inside, held while its declarations are read
+ * so each can be given the path it sits at.
+ *
+ * A skipped module is on the stack for its indent alone, so that what it
+ * contains is dropped with it rather than losing the enclosing segment from its
+ * path. Nothing is ever pushed above a skipped entry, because every declaration
+ * under one is skipped before it can be.
+ */
+interface OpenModule {
+    /** Dotted path of the module itself, enclosing segments included. */
+    name: string;
+    indent: number;
+    skipped: boolean;
+    /**
+     * True only between a module declaration that ended without opening a body
+     * and the next line considered, which is where its `{` may be.
+     */
+    awaitingBrace: boolean;
+    /**
+     * Brace depth just outside a braced body, and null while indentation is
+     * what closes the entry: the colon, dotted and `>` forms, and a braced one
+     * whose `{` is still to come.
+     */
+    closeDepth: number | null;
+}
 
 /**
  * Reads the project's .verse files into a flat list of the module, class,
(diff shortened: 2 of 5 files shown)


vukefn:
Filed the pre-existing defect found while reviewing this change as #342. It is a different route to the harm class of #330 and is untouched by this PR.

PR #343, opened 2026-08-12
fix(services): bound a braced body opened inside a skipped module

### LINKED ISSUE #342: A braced body inside a skipped module is untracked, so its modules are offered at file scope
Found by the review gate on #336, which fixed a different route to the same harm. Present on `main` before that change and unaffected by it.

## What happens

A module inside a skipped module's braced body escapes the drop and is offered as a top-level import candidate.

```verse
Priv<private> := module:
    Inner := module{
X := module {}
    }
Y := module {}
```

`ProjectPathScanner.extractDeclarations` returns `["X", "Y"]`. `X` is really `Priv.Inner.X`, inside a private module, so it is not importable at any path — and `X` collides with any genuine top-level module of that name.

## Why

`src/services/ProjectPathScanner.ts:298` skips the rest of the line as soon as the stack top is skipped, and that `continue` sits ahead of every push:

```ts
if (moduleStack.length > 0 && moduleStack[moduleStack.length - 1].skipped) {
    continue;
}
```

So nothing inside a skipped module is ever pushed. The braced `Inner` body is therefore untracked — no entry carries its `closeDepth` — and the only entry on the stack is the skipped colon `Priv`, which indentation closes at `X`, where `0 <= 0`. From there the scan believes it is at file scope, and `X` is recorded with no prefix.

The skip is deliberate and its reasoning is sound for the colon form: everything under a skipped module is as unreachable as the module, so recording the subtree without its enclosing segment would offer a path that does not exist. The gap is that a braced body opened *inside* the skipped subtree still has to be tracked for the depth to close correctly, even though nothing in it is recorded.

## Note on the fix

#336 added the two-phase close, so the stack can now be truncated from below by brace depth. That machinery is what a fix here would build on: a skipped module needs its interior brace structure tracked (entries pushed but marked skipped, or a depth counter for the skipped span) so the scan does not return to file scope early.

Whatever the shape, the acceptance criterion is that the input above yields no nodes at all, and that a sibling declared after the skipped module's body still resolves at file scope.

## Coverage

- The input above, asserting no declarations.
- The same with the skipped module written `<scoped{ModuleA}>` rather than `<private>`.
- A kept sibling after the skipped body, which must still be top-level.
- The colon-only form, which already drops correctly today and must keep doing so.

vukefn:
Closes #342

A module declared inside a skipped one was never pushed onto the module stack, so a braced body opened in there had no entry carrying the depth its `}` closes at. The skipped entry above it is delimited by indentation, and a braced body's members may be written back at that indent, so the first one popped it — and the scan carried on believing it was at file scope while still inside the subtree.

## What changed

- `src/services/ProjectPathScanner.ts` — `skipped` is now inherited. The stack-top-is-skipped test becomes a read (`insideSkipped`) instead of a `continue`, so a module declared inside the subtree is still pushed, marked skipped in turn; every declaration below the module branch is dropped where it stands. The two `moduleStack.push` calls collapse into one. The `OpenModule` doc drops its "nothing is ever pushed above a skipped entry" invariant, which this change replaces.
- Tests — the ticket's four shapes, a next-line-brace variant, and the two the review gate asked for.

`outermostClosedBrace` and `closedByIndent` needed no change: a skipped entry now carries the same `closeDepth`/`indent` as a kept one, so #336's two-phase close bounds it without knowing the difference.

## One correction to the ticket

The acceptance sentence says the input "yields no nodes at all". That cannot hold alongside the same sentence's second half, "a sibling declared after the skipped module's body still resolves at file scope" — `Y := module {}` in that very input *is* that sibling. It sits at indent 0, back at `Priv`'s own indent, so it dedents out of `Priv`'s colon body and is a genuine top-level module; dropping it would be a new defect. The ticket's own "Why" names only `X` as wrong.

The implemented expectation is therefore `["Y"]`, not `[]`. The review gate was asked to judge this independently and reached the same conclusion on three grounds, confirming `origin/main` returns exactly `["X", "Y"]`.

## Verification

```
npm run compile
> tsc -p ./
(clean)

npm test
Tests:       977 passed, 977 total

npm run format:check
```

The five ticket-shaped tests were run against the unfixed scanner first: four failed with exactly the stray top-level `X` the ticket predicts, and the fifth — the colon-only pin — passed, which is what the ticket says it should do. The two tests added after review also fail on `origin/main`.

## Review

One round, verdict PASS_WITH_SUGGESTIONS, no Critical and no Important. The reviewer probed rather than only read: 40-odd shapes across all four dropped specifiers, `includePrivate` on each, kept-parent controls compared byte-for-byte against `origin/main`, deep brace/colon interleavings, stray braces, a braced `using` clause and a `'{'` char literal inside the skipped subtree, `<#>` comments, CRLF and tab input, next-line-brace timing, all seven non-module declaration forms, and two mutation tests against the new guards. Every kept path and every `includePrivate` path it probed is identical to `origin/main`.

Five of its six suggestions are applied here:

- **Kept-parent coverage gap.** The fix repairs a second harm nothing pinned: with a kept enclosing module, `origin/main` returned `["Keep", "X", "Sib", "Y"]` — the leaked `X`, *and* a kept sibling stripped of its `Keep.` prefix. Now pinned.
- **Unclosed-brace degradation.** An unclosed body inside the skipped subtree now costs the declarations after it where `main` offered them at a path that does not exist. It does not compile either way, and the kept path already behaves identically. Pinned as degradation, in the spirit of the existing "a module missing its body cannot swallow the rest of the file" test.
- The `insideSkipped` comment restated the `OpenModule` doc nearly clause for clause; trimmed to what is local to the read.
- A test name claimed the colon-only form closes "without any entry of its own", which this change makes false.
- The changelog fragment named only private and scoped; `<protected>` and `<epic_internal>` drop through the same code and had the same leak.

The sixth is out of scope and left unapplied: `block:`, `if:` and `loop:` inside a function body match `variablePattern` and are recorded as junk `variable` nodes in the declaration cache. It reproduces identically on `origin/main` and is unrelated to this ticket — worth its own issue.

### DIFF
--- a/changelog.d/342.fixed.md
+++ b/changelog.d/342.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: A module inside a private or scoped module's braced body is no longer offered as a top-level import candidate, where it named a path the compiler rejects and collided with any real module of that name.
--- a/changelog.d/342.fixed.md
+++ b/changelog.d/342.fixed.md
@@ -1,1 +1,1 @@
-**Project scan**: A module inside a private or scoped module's braced body is no longer offered as a top-level import candidate, where it named a path the compiler rejects and collided with any real module of that name.
+**Project scan**: A module inside the braced body of a module dropped for its access level is no longer offered as a top-level import candidate at a path the compiler rejects.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -304,11 +304,9 @@ export class ProjectPathScanner {
             // and is dropped with it. The pop loop above has already closed
             // the module for a line that left it.
             //
-            // Read rather than acted on here, because a module declared inside
-            // the subtree still has to reach the stack: only an entry of its
-            // own carries the depth its braced body closes at, and without one
-            // the enclosing skipped entry is popped by the first member written
-            // back at its indent, which a braced body permits.
+            // Read rather than acted on, because a module declared in there
+            // still has to reach the stack to bound its own body; only the
+            // declarations below it are dropped where they stand.
             const insideSkipped = moduleStack.length > 0 && moduleStack[moduleStack.length - 1].skipped;
 
             currentModulePath = moduleStack.length > 0 ? moduleStack[moduleStack.length - 1].name : "";
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -11,10 +11,12 @@ const SCAN_CONCURRENCY = 8;
  * A module whose body the scan is inside, held while its declarations are read
  * so each can be given the path it sits at.
  *
- * A skipped module is on the stack for its indent alone, so that what it
+ * A skipped module is on the stack for its bounds alone, so that what it
  * contains is dropped with it rather than losing the enclosing segment from its
- * path. Nothing is ever pushed above a skipped entry, because every declaration
- * under one is skipped before it can be.
+ * path. A module declared inside one is pushed too, marked skipped in turn: a
+ * braced body opened in there closes on its brace, and an entry no one tracks
+ * leaves that depth unaccounted for, which returns the scan to file scope while
+ * it is still inside the subtree.
  */
 interface OpenModule {
     /** Dotted path of the module itself, enclosing segments included. */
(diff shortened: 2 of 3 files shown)


vukefn:
Filed the pre-existing block-macro finding from this PR's review gate as #344.

PR #346, opened 2026-08-12
fix(imports): keep a path conversion off a using that names a module alias

### LINKED ISSUE #71: No awareness of import(...) module aliases
## Problem

The Book of Verse documents the `import` expression for module aliases ([Module Aliases with import](https://verselang.github.io/book/16_modules/#module-aliases-with-import)):

```verse
Utils := import(/MyProject/Utilities)
Result := Utils.HelperFunction()

# Combined form: alias plus direct member access
Graphics := import(/MyProject/Systems/Graphics)
using { Graphics }
```

The extension has no awareness of this construct (no handling anywhere in src/). Consequences:

- Dedupe gap: auto-import can add `using { /MyProject/Utilities }` for a module the file already aliases.
- The combined form's `using { Graphics }` line is a bare identifier and is treated as local-scope using (see #65).
- Optimize Imports ignores alias lines entirely - probably correct, but currently an accident rather than a decision.

## Ask

Decide the intended support level and document it. Minimum viable: treat `X := import(/Path)` as evidence that `/Path` is already imported (dedupe), and never touch alias lines during optimize. Add fixtures and tests for alias-containing files so future changes do not regress them.

Before implementing, validate that `import(...)` is accepted by the live UEFN compiler (the book may be ahead of the shipped toolchain); record results in the test corpus.

Source: systematic comparison of import handling against the Book of Verse (2026-07-03).

vukefn:
Closes #71

## What this is

A file can bind a name to a module with `Name := import(/Path)`. A `using` of that name is an ordinary identifier, indistinguishable by content from a same-directory folder-module import — so `scanConvertibleImports` offered a path-conversion CodeLens on it, and `ImportPathConverter` resolved the name by searching the workspace for a folder of that name or a `:= module` declaration of it. Whichever namesake it found got written into the line, silently importing a different module than the alias names. "Use absolute paths for all" applied it in bulk, without the user looking at that line.

`scanModuleAliases` now reads the names a file declares, and the shared convertible set drops a `using` rooted at one of them. That set is the one membership function behind both the lens provider and the converter, so a single filter closes both entry points.

## What the ticket asked for that this deliberately does not do

The body's "minimum viable" ask was to treat `X := import(/Path)` as evidence that `/Path` is already imported, for dedupe. **That is not built, because it is wrong**, as the ticket's own third comment records: aliasing does not bring the module's members into scope, so an alias never substitutes for a `using`, and a compiler suggestion to add `using { /A/B }` stays correct beside an alias for `/A/B`. Building the dedupe would decline an import the file needs.

That decision is encoded in the API rather than left in prose: `scanModuleAliases` returns the alias *names* only and never the paths they alias, so nothing downstream can read a path out of it and treat it as an import already present.

## Domain rules this rests on, with citations

From the compiler source (`EpicGames/UnrealEngine`, `ue5-main`, `Engine/Source/Runtime/VerseCompiler`):

- `AnalyzeImport` (`SemanticAnalyzer.cpp:19396-19462`) fixes the statement shape: module scope only, parentheses not brackets (brackets carry failure semantics), LHS a plain identifier with no qualifier and no type annotation, and the argument a path literal — so an alias always names an absolute path. `MODULE_ALIAS_STATEMENT` matches that shape and nothing wider.
- The alias is a real `CModuleAlias` definition (`:19446`) and registers no using scope, which is the citation behind not building the dedupe.
- Hoisting a `using { Alias }` above the alias line is safe: `Deferred_Import` runs to completion strictly before `Deferred_ModuleReferences` (`:19453` enqueues alias resolution into the former; `:21304-21305` runs the stages in that order), so every alias is bound before any module-scope `using` is resolved. Confirmed a second way: the `using` handler analyzes its argument with `WithResultIsDotted` (`:6567`), which is what the `CModuleAlias` guard at `:12037` requires, so the reference resolves whichever order the two lines are written in.

Live-compiler status: the construct compiles in real UEFN 41.10 (the maintainer's smoke run on this issue, PLAYBOOK T9 case F).

## Known limit, and where it came from

**An alias binds in the module, not the file.** A snippet is not a logical scope (`SemanticProgram.h:185-192`), a module part reports the `CModule` as its logical scope (`:173-174`), and `CScope::CreateModuleAlias` adds to `GetLogicalScope()` (`SemanticScope.cpp:437-442`) — so a sibling `.verse` file's alias resolves here unqualified, and Epic's own corpus has an `assert_valid` titled "Importing something in one snippet is also visible in another".

This change reads one file, so a `using` naming an alias declared in a sibling file is still offered a conversion. That is the same defect, narrower. Closing it needs an async workspace scan that this line-based reader cannot make — a different shape of change, touching the converter and the lens provider's async paths. `scanModuleAliases` states the limit rather than denying it. Worth a follow-up issue.

Two claims the review pushed back on are now stated at the strength the source supports rather than asserted: the compiler rejects the bracketed `import[...]` through a failure-semantics check that never inspects the brackets, and the `(path:)` qualifier on the left side is not checked at all. Both are now **read** rather than refused, because over-reporting a name only withholds a lens while missing one leaves the defect live.

## What was already correct, now pinned

Organize left alias lines byte-identical before this change — the ticket calls that "an accident rather than a decision". Four tests at the `buildOrganizedContent` entry point make it a decision: an alias between two imports stays in the body written as it was, a comment above an alias travels with the alias, an import of a path the file only aliases is still added, and the combined form survives with both lines intact.

## Fixtures

`test-fixtures/uefn-smoke/Content/T9/t9_alias.verse` was a commented-out probe asking whether the construct compiles. That question is settled, so the file is now written live and carries all three cases — a plain alias, the combined form, and a `using` of a path the file also aliases. The PLAYBOOK T9 case F step is rewritten from "uncomment and see if it compiles" into the three live checks the fixture now supports.

## Verification

```
$ npm run compile
> tsc -p ./
```

```
$ npm test
> jest --verbose

Tests:       993 passed, 993 total
Snapshots:   0 total
Time:        9.893 s, estimated 16 s
Ran all test suites.
```

```
$ npm run format:check
> prettier --check "**/*.ts"

Checking formatting...
```

The two exclusion tests were confirmed to detect the defect: with the alias filter removed from `scanConvertibleImports` and everything else in place, exactly those two fail.

### DIFF
--- a/changelog.d/71.fixed.md
+++ b/changelog.d/71.fixed.md
@@ -1,1 +1,1 @@
-**Path conversion**: A `using` naming a module alias the file declares with `import(...)` no longer gets a path-conversion lens, so converting paths one at a time or in bulk cannot rewrite it into an import of an unrelated module of the same name.
+**Path conversion**: A `using` naming a module alias the file declares with `import(...)` no longer gets a conversion lens, so converting paths singly or in bulk cannot rewrite it into an import of an unrelated namesake module.
--- a/changelog.d/71.fixed.md
+++ b/changelog.d/71.fixed.md
@@ -1,0 +1,1 @@
+**Path conversion**: A `using` naming a module alias the file declares with `import(...)` no longer gets a path-conversion lens, so converting paths one at a time or in bulk cannot rewrite it into an import of an unrelated module of the same name.
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -1190,6 +1190,88 @@ export function allUsingPaths(lines: string[]): string[] {
     return paths;
 }
 
+/**
+ * One `Name := import(/Path)` module alias, anchored at the head of a statement.
+ *
+ * The shape is the compiler's, not a generalization of it: the left side is a
+ * plain identifier, optionally carrying a quoted suffix and access specifiers,
+ * and the argument is a single absolute path literal in parentheses. A
+ * qualified left side, a type annotation, brackets rather than parentheses, and
+ * a relative or dotted argument are each rejected outright by the compiler, so
+ * a line spelling one declares no alias and must not be read as if it did.
+ *
+ * The path is captured but deliberately not offered by scanModuleAliases; see
+ * there for why nothing may read it.
+ */
+const MODULE_ALIAS_STATEMENT = /^([A-Za-z_][A-Za-z0-9_]*(?:'[^']*')?)(?:\s*<[^<>\n]+>)*\s*:=\s*import\s*\(\s*(\/[^\s()]*)\s*\)$/;
+
+/**
+ * The names a file binds to a module with `Name := import(/Path)`, at file
+ * scope.
+ *
+ * The names alone, never the modules they alias. An alias does not bring its
+ * module's members into scope - it binds one name to one module - so a file
+ * that aliases `/A/B` still needs `using { /A/B }` to use that module's
+ * members unqualified, and a diagnostic asking for one is right to. Offering
+ * the aliased path here is what invites a reader to treat it as an import
+ * already present, which suppresses an import the file needs.
+ *
+ * What the names are for is the opposite question: which `using` statements
+ * name something that is not a module path. An alias is an ordinary identifier
+ * bound in the file's own scope, so `using { Gfx }` naming an alias is
+ * indistinguishable, by content alone, from one naming a same-directory folder
+ * module - and a reader that resolves it as a path finds whichever namesake the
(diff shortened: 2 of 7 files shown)


vukefn:
Filed #347 for the cross-file residual: an alias binds in the enclosing module rather than the file, so a `using` naming an alias declared in a sibling `.verse` file of the same folder module is still offered a conversion. Closing it needs workspace-aware alias discovery, which this line-based reader cannot do; `scanModuleAliases` documents the limit in the meantime.

Issue #344, opened 2026-08-12
A Verse block macro is recorded as a variable declaration by the project scan


vukefn:
Found by the review gate on #342, which fixed something else in the same scan. Present on `main` before that change and unaffected by it.

## What happens

A Verse block macro written inside a function body is recorded as a `variable` declaration in the project scan's cache.

```verse
Inventory := module:
    Run<public>()<suspends>:void =
        block:
            X:int = 1
        loop:
            Sleep(1.0)
        if:
            true
        then:
            Print(:)
        race:
            Sleep(1.0)
        sync:
            Sleep(1.0)
        branch:
            Sleep(1.0)
        defer:
            Print(:)
```

`ProjectPathScanner.extractDeclarations` returns, alongside the two real declarations:

```
variable:Inventory.block
variable:Inventory.loop
variable:Inventory.if
variable:Inventory.then
variable:Inventory.race
variable:Inventory.sync
variable:Inventory.branch
variable:Inventory.defer
```

None of these is a declaration and none of them names anything. They are keywords, so the cache gains one junk node per block macro per function across the whole project.

## Why

`src/services/ProjectPathScanner.ts:231` is the fall-through pattern, and it needs only a name followed by a colon:

```ts
const variablePattern = /^(\w+)((?:<[^>]+>)*)\s*:/;
```

A bare `block:` satisfies that. Its branch at `:487` carries two guards, and the comment above them says what they are for — a type recognised by `:=` with a declaration keyword, and a function recognised by a parameter list before the colon. Both are backstops for a *declaration* whose specific pattern did not match the exact shape written. A block macro is not a declaration at all, so neither guard is aimed at it and neither fires.

The scan does not track function bodies — the class doc says so, and only module nesting is tracked — so there is no enclosing-scope test available to reject these on position. The keyword set is what separates them.

## Why it matters

The nodes are offered downstream as import candidates and completion entries. `Inventory.block` is not importable, and the names are common enough across a project that they crowd real candidates: one `Inventory.if` per module with a function in it.

Severity is low — a user is unlikely to accept `if` as an import — so this is cache pollution and noise rather than a wrong edit.

## Note on the fix

[...]

PR #352, opened 2026-08-12
fix(services): reject a parenthesised Verse block macro as a declaration

### LINKED ISSUE #350: A parenthesised Verse block macro is recorded as a function declaration by the project scan
Found by the review gate on #344, which fixed the bare-colon spelling of the same problem in the branch below this one. Present on `main` before that change and unaffected by it.

## What happens

A Verse block macro written with a parenthesised head inside a function body is recorded as a `function` declaration in the project scan's cache.

```verse
Inventory := module:
    Run<public>()<suspends>:void =
        if (X > 0):
            Print(:)
        for (Item : Items):
            Print(:)
        case (X):
            Print(:)
```

`ProjectPathScanner.extractDeclarations` returns, alongside the two real declarations:

```
function:Inventory.if
function:Inventory.for
function:Inventory.case
```

## Why

`src/services/ProjectPathScanner.ts` reaches `functionPattern` before the fall-through variable pattern:

```ts
const functionPattern = /^(\w+)((?:<[^>]+>)*)\s*\([^)]*\)\s*(?:<[^>]+>)*\s*:/;
```

`if (X > 0):` satisfies it - a name, a parenthesised list, a colon. The branch has no guard of its own; it is the specific pattern the fall-through's guards are backstops *for*.

#344 added `RESERVED_LINE_KEYWORDS` and a guard rejecting a fall-through match whose name is one of those keywords. That guard sits in the variable branch alone, so it never sees a line the function branch has already claimed.

## Why it matters

`if (Cond):` and `for (X := 0..10):` are the dominant spelling of these macros in real Verse - considerably more common than the bare `if:` that #344 sampled. The nodes are offered downstream as import candidates and completion entries, and `Inventory.if` is not importable.

Severity is low for the same reason #344's was: a user is unlikely to accept `if` as an import, so this is cache pollution and noise rather than a wrong edit. It is worth noting that the junk is typed `function` here, so a cleanup aimed at bogus `variable` nodes would leave it in place.

## Note on the fix

`RESERVED_LINE_KEYWORDS` already exists and already holds every keyword involved, each cited to the compiler's `ReservedSymbols.inl` or to the parser token table in `VerseGrammar.h`. The smallest change is to apply the same membership check in the `functionPattern` branch.

Worth deciding deliberately whether the check belongs in each branch or once, ahead of all of them: a reserved word cannot name any declaration, so the rule is not specific to either branch, and the extension method branch above takes a name from a different capture group and would need its own handling either way.

## Coverage

- The input above, asserting the three junk nodes are gone and both real declarations survive.
- A real function whose name merely starts with a keyword (`ifMatches(X:int):logic = ...`), which must still be recorded.
- The bare-colon cases #344 pinned, unchanged.

vukefn:
Closes #350

`functionPattern` in `ProjectPathScanner.extractDeclarations` needs a name, a parenthesised list and a colon, which the parenthesised spelling of a Verse block macro satisfies. `if (X > 0):`, `for (Item : Items):` and `case (X):` were each recorded as a `function` declaration and offered downstream as an import and completion candidate — and typed `function`, so the cleanup #344 aimed at bogus `variable` nodes left them in place.

#344's `RESERVED_LINE_KEYWORDS` guard sits on the variable fall-through alone, so it never sees a line the function branch has already claimed.

## The fork the ticket raised

The ticket asked for a deliberate choice between a per-branch check and one ahead of all branches. This takes the hoisted form: a reserved word cannot name *any* declaration, so the rule is not specific to a branch, and a block macro reaches two different patterns depending on how it is written — the variable fall-through for `keyword:`, the function pattern for `keyword (Cond):`. A guard on either branch alone leaves the other spelling recorded, which is how the bare-colon fix left this one open.

The guard replaces the variable branch's copy, so the rule is stated once.

## The rule this relies on, and its authority

A word Verse reserves cannot name a declaration.

- `case` and `for` are `EIsReservedSymbolResult::Reserved` in `VerseCompiler/Public/uLang/Parser/ReservedSymbols.inl:25,59`.
- `if` is absent from that file by design — its header says language-level reservations live in the parser token list instead — and is in `VerseGrammar.h`'s `Tokens[]` table at line 674.
- The one demotion of a reserved word to an identifier is `VerseGrammar.h:1174-1189`, and it requires `:=` to follow. That is an object-notation key rather than a declaration, and the type patterns all require `:=` plus a declaration keyword, so hoisting the guard above them cannot drop a legal declaration. The extension-method branch has no leading `\w+` at all, since its line opens with `(`.
- The parenthesised heads are the real spelling: `if (…):`, `for (Item : Items):` and `case (Dir):` throughout the Book of Verse, and `VerseGrammar.h:2744` comments on `if(a)` at the grammar level.

## Changes

- `src/services/ProjectPathScanner.ts` — a `leadingWordPattern`, and a guard above the pattern dispatch rejecting a line whose leading word is in `RESERVED_LINE_KEYWORDS`. The variable branch's copy is removed. Placement is load-bearing in both directions and the comment says so: below the brace counting and module-stack maintenance, above the module push.
- `src/services/__tests__/ProjectPathScannerBlockMacros.test.ts` — the ticket's exact input, nine parenthesised heads including the no-space `if(X > 0):`, and functions whose names merely start with a keyword (`ifMatches`, `forEach`, `caseOf`). The bare-colon cases #344 pinned are unchanged.
- `changelog.d/350.fixed.md`.

The tests were confirmed to detect the defect: written first, they failed 9 for 9 against the unfixed code, each reporting the junk node the ticket names (`function:Inventory.if`, `function:Inventory.loop`, …).

## Verification

`npm run compile`

```
> verse-auto-imports@0.11.0 compile
> tsc -p ./
```

`npm test`

```
> verse-auto-imports@0.11.0 test
> jest --verbose

Test Suites: 41 passed, 41 total
Tests:       1049 passed, 1049 total
Snapshots:   0 total
Time:        13.63 s
Ran all test suites.
```

`npm run format:check`

```
> verse-auto-imports@0.11.0 format:check
> prettier --check "**/*.ts"

Checking formatting...
```

## Review

One round, fresh Opus reviewer, verdict `PASS_WITH_SUGGESTIONS`, 0 Critical, 0 Important.

It attacked the hoist's safety argument with a 75-input old-versus-new differential — 30 synthetic sources, all 43 `test-fixtures/*.verse`, and the three bundled digests — and found 68 identical, with all 7 differences pure removals of junk nodes and no legal declaration lost. It also probed every block-macro spelling, keyword-prefixed real declarations, braced and nested module bodies, extension methods, skipped private modules, CRLF, tabs, comments and string literals holding keywords, uppercase spellings, and specifiers.

Two of its four suggestions are applied here: the comment stated only one half of the guard's load-bearing placement, and the suite did not pin the no-space spelling.

One is deliberately not: it trimmed further comment argumentation, and the remaining text is the constraint rather than the defence.

Its last is out of this ticket's scope — `array` and `map` are `:`-headed macros absent from `RESERVED_LINE_KEYWORDS`, so `array:` on its own line is still recorded as a variable. Identical on both sides of this diff. Worth filing separately.

### DIFF
--- a/changelog.d/350.fixed.md
+++ b/changelog.d/350.fixed.md
@@ -1,0 +1,1 @@
+**Project scan**: a Verse block macro written with a parenthesised head, such as `if (Cond):` or `for (Item : Items):`, is no longer recorded as a function declaration or offered as an import candidate.
--- a/src/services/ProjectPathScanner.ts
+++ b/src/services/ProjectPathScanner.ts
@@ -9,12 +9,12 @@ const SCAN_CONCURRENCY = 8;
 
 /**
  * The keywords a Verse block macro is spelled with, and the operators and types
- * that share their shape closely enough to reach the same fall-through.
+ * that share their shape closely enough to reach a declaration pattern.
  *
- * Membership is not "every reserved word": it is the words seen to reach the
- * fall-through as a bare `keyword:`, plus the ones the language groups with
- * them. What makes rejecting them safe is that Verse reserves each, so none can
- * name a declaration - the parser demotes a reserved word to an identifier only
+ * Membership is not "every reserved word": it is the words seen to open a line
+ * a declaration pattern matches, plus the ones the language groups with them.
+ * What makes rejecting them safe is that Verse reserves each, so none can name
+ * a declaration - the parser demotes a reserved word to an identifier only
  * where `:=` follows it, and that is an object-notation key rather than a
  * declaration either way. `block`, `loop`, `race`, `rush`, `sync`, `branch`,
  * `defer`, `spawn`, `case`, `for`, `and`, `or`, `option` and `logic` are
@@ -248,6 +248,8 @@ export class ProjectPathScanner {
         /** `(Type:type).Name<specifiers>(params)<effects>:` */
         const extensionMethodPattern = /^\([^)]+\)\.(\w+)((?:<[^>]+>)*)\s*\([^)]*\)/;
         const variablePattern = /^(\w+)((?:<[^>]+>)*)\s*:/;
+        /** The word a line opens with, absent where it opens with anything else. */
+        const leadingWordPattern = /^(\w+)/;
 
         for (let i = 0; i < lines.length; i++) {
             const line = lines[i].trim();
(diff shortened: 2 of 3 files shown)


vukefn:
The reviewer's remaining out-of-scope finding is filed as #353: `array` and `map` are `:`-headed macros absent from `RESERVED_LINE_KEYWORDS`, so a bare `array:` is still recorded as a variable. Identical on both sides of this diff.

Issue #370, opened 2026-08-13
Logger: migrate display to LogOutputChannel, keep the ring buffer and export


vukefn:
## What

Replace the Logger's two plain output channels with a single `LogOutputChannel` for display, and keep the ring buffer and the export command exactly as they are. A full migration is the wrong move: it would destroy the post-hoc export.

## Where

`logger.ts` internals — the ring buffer at `logger.ts:32, 258-263`, the unconditional buffering in `Logger.log` at `:236-263`, and roughly 80 lines of formatting the migration deletes. The two channel consumers that change: `extension.ts:74` and `StatusBarHandler.ts:287`. Tests under `src/utils/__tests__/` are affected.

## Effect

Today the extension carries its own formatting and two Output entries, and gets no standard "Developer: Set Log Level" support. Migrating display to `LogOutputChannel` buys that support and one Output entry, and deletes the formatting code.

The constraint that decides the shape: `LogOutputChannel` filters at write time by the user's chosen log level, while `Logger.log` buffers every level unconditionally and the user channel filters at INFO. The buffer therefore captures TRACE and DEBUG detail from before the user knew they had a problem, which is the whole point of a post-hoc export command. A naive full migration drops those entries at write time and destroys it.

## Evidence

Re-verified in source: unconditional buffering at `logger.ts:236-263` against an INFO-filtered user channel, and the write-time filtering behavior of `LogOutputChannel`. Call sites already go through the singleton, so only `logger.ts` internals and the two channel consumers change.

## Smallest fix

One `LogOutputChannel` for display; keep the ring buffer and the export command. The strongest argument against acting at all, recorded for the record: the logger works, the export is the extension's flagship diagnostic, and the gain is cosmetic.

---
**Severity:** A-J1 is a judgment row and carries no severity label; the ruling is that the partial migration survives.
**Source:** lane A platform review, row A-J1. Full finding in `docs/review/lanes/A-platform/02-findings.md`, re-derived in `03-rulings.md`.

vukefn:
Closing as deferred, not rejected, on the review campaign's synthesis (lane D) and its gate.

Both passes agree the churn exceeds the value at present. One correction to the synthesis's reasoning, from its gate: the ticket as filed is already the hazard-avoiding partial migration — it keeps the ring buffer and the export command, so it does not carry the risk of losing retroactive TRACE/DEBUG detail that a naive full migration would. The deferral rests on cost and benefit alone, not on a risk in the ticket.

[...]

Issue #369, opened 2026-08-13
Deprecate general.diagnosticDelay via deprecationMessage


vukefn:
## What

`general.diagnosticDelay` is deprecated in prose inside its `description` instead of via `deprecationMessage`, so the settings UI still presents it as a first-class setting above its replacement.

## Where

`package.json:165-170` — the prose deprecation inside `description`. The property sits at order 2, above its replacement at order 3. No property in the manifest uses `deprecationMessage`.

## Effect

The settings UI shows a deprecated key first and its replacement second, actively inviting new users to set the wrong key. VS Code's deprecation support would hide it from the UI and flag existing values.

## Evidence

Manifest read and re-verified on the rulings pass: the prose-only deprecation, the order 2 against order 3 placement, and the absence of `deprecationMessage` on any property.

## Smallest fix

Move the deprecation from prose into a `deprecationMessage` on the property.

---
**Severity:** verdict item 7b carries no severity label in the source.
**Source:** lane A platform review, verdict item 7b. Full finding in `docs/review/lanes/A-platform/02-findings.md`, re-derived in `03-rulings.md`.

Issue #368, opened 2026-08-13
Inject one ImportPathConverter instead of constructing a second


vukefn:
## What

Two `ImportPathConverter` instances with different capabilities sit on either side of one user interaction: activation builds a cache-aware converter for the command side, and `ImportCodeLensProvider` self-constructs a second one without the cache.

## Where

`extension.ts:92` builds the command-side converter with `projectPathCache`; `ImportCodeLensProvider.ts:25` constructs its own without it. Each converter also builds its own `ProjectPathHandler` (`ImportPathConverter.ts:62`), so the count of duplicated collaborators is growing. The self-construction is the only place activation's deliberate DI wiring (`commandsDeps`, `extension.ts:96-103`) is bypassed.

## Effect

No divergence is observable today: the provider calls exactly three methods on its converter — `isFullPathImport` (`ImportPathConverter.ts:167-170`), `isBuiltinModule` (`:180-183`) and `extractModuleName` (`:236-243`) — and all three resolve through `extractPathFromImport` into pure string work that never touches `projectPathCache`. The cost is standing risk plus precedent. The moment any lens-side judgment consults the cache-aware path — the natural next feature, "only offer the lens when the conversion will succeed" — the lens and the command it invokes will answer the same question differently, and the failure will be a lens that offers a conversion the command then refuses (`CommandsHandler.ts:493`).

## Evidence

Both construction sites read. The three provider-invoked methods were traced and confirmed cache-free. Nothing upgrades the provider's converter after construction either: `setProjectPathCache` (`ImportPathConverter.ts:71-73`) has no callers anywhere in `src/`. The refusal path the item predicts exists at `CommandsHandler.ts:493`.

## Smallest fix

One constructor parameter: pass the existing cache-aware converter into `ImportCodeLensProvider` instead of letting it self-construct a second one.

---
**Severity:** verdict item 4 carries no severity label; it is recorded as standing risk plus precedent, not a live bug, and that framing was verified rather than assumed.
**Source:** lane A platform review, verdict item 4. Full finding in `docs/review/lanes/A-platform/02-findings.md`, re-derived in `03-rulings.md`.

Issue #367, opened 2026-08-13
Delete dead state: refreshTimeout, setProjectPathCache, and two inaccurate comments


vukefn:
## What

Two pieces of state exist that nothing uses, and two comments describe something other than the code they sit above. Each costs a reader who has to prove the absence for themselves.

## Where

- `ImportCodeLensProvider.ts:21` — `refreshTimeout`, cleared at `:39-42` and `:253-256`, armed nowhere.
- `ImportPathConverter.ts:71-73` — `setProjectPathCache`, with no callers anywhere in `src/`.
- `CommandsHandler.ts:61-63` — the registration doc comment says "the five that take caller-supplied arguments must stay hidden"; six commands are hidden and take arguments (`addSingleImport`, the four convert commands, `makeModulePublic`).
- `StatusBarHandler.ts:65-69` — the snooze comment's blanket main clause: "Persisting it (or mirroring it into user settings) is what could leave auto imports off forever after a reload mid-snooze."

## Effect

Every reader of `ImportCodeLensProvider` searches for the assignment that does not exist and concludes they misread the file. `setProjectPathCache` is dead public API whose existence invites "fixing" the duplicate-converter problem per call site instead of via constructor injection. The registration comment undercounts the very set it exists to protect, which is the failure mode a protective comment cannot afford. The snooze comment's main clause is false for the design it is read against.

## Evidence

`refreshTimeout` declared, cleared twice, armed nowhere — verified on both passes. `setProjectPathCache` has no callers in `src/` — verified during the item 4 re-derivation. Six hidden argument-taking commands are pinned at `package.json:129-154` with `"when": "false"`, and each hidden handler's signature dereferences its arguments (`CommandsHandler.ts:103, 117, 484, 521, 561, 582`).

On the snooze comment specifically: it is not a strawman in origin. It first appears in `41623ae` ("fix(ui): hold snooze state in memory instead of user settings", closes #132), whose message records the real incident — snooze wrote `general.autoImport: false` into global settings, relied on an interval to write it back, and a reload mid-snooze left auto import permanently off. The comment argues correctly against a settings mirror or a persisted flag. What is false is the blanket clause: a persisted end-timestamp, read and expired at activate, expires by wall clock and cannot outlive its own deadline by construction, so the comment does not answer that proposal at all.

## Smallest fix

[...]

PR #371, opened 2026-08-13
docs: add DECISIONS.md

vukefn:
Adds `DECISIONS.md`: choices this project has made on purpose and does not want raised again.

Ten entries, from a lane-by-lane review of the extension. Each records the decision, why it holds, and — the part that matters most — the condition that would make it wrong. A decision with no expiry condition is a habit, not a decision.

Some examples of what it settles:

- Keep the `engines.vscode` floor at `^1.85.0`. The newest API the extension uses is 1.67-era, so raising it buys zero capability today and drops users the Marketplace still serves.
- Keep snooze state in memory. It carries the incident behind it: snooze once wrote `autoImport: false` into global settings, and a reload mid-snooze left auto-import off permanently.
- Keep the single-file alias scan boundary. A review challenged its justification as overstated; the challenge was wrong, and `Import.versetest:58-60` is why.
- Digest data reaches the repository through pull requests. A review read `[skip ci]` in `parse-digest.yml` and filed it as an open gate. That workflow's push step has never succeeded and cannot — the token is read-only and a ruleset requires a PR.

That last one is the pattern worth keeping: a refuted finding is worth recording precisely *because* it looked true. The next reader will see the same `[skip ci]` and reach the same wrong conclusion unless something explains why it cannot fire.

The review's working record — inventories, rubrics, findings and the refute rulings that re-derived them — is kept locally under `.claude/review/` rather than in this repository. Every entry here is written to stand on its own, so a reader without those files loses the working, not the decision.

Documentation only. No source changes.

### DIFF
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -1,0 +1,119 @@
+# Decisions
+
+Choices this project has made on purpose, and does not want raised again.
+
+A review that reaches one of these should stop. If the reasoning below no longer holds, that is worth a ticket — say what changed. If it still holds, there is nothing to file.
+
+Each entry records the decision, why it holds, and what would make it wrong. The last part matters most: a decision with no expiry condition is a habit, not a decision.
+
+## Keep the `engines.vscode` floor at ^1.85.0
+
+Raising the floor drops every user on an editor below it, because the Marketplace serves them the last compatible version instead. The lane A review checked every VS Code API this extension uses: the newest is 1.67-era (`window.tabGroups`, `TabInputText`). Raising the floor today buys zero capability.
+
+**Revisit when** a wanted API postdates 1.85. Then the trade is real and worth pricing.
+
+Note that the floor is currently a promise no tool enforces, because `@types/vscode` is declared with a caret and floats far above it. That is a defect, not a decision, and it has its own ticket.
+
+## Keep snooze state in memory
+
+Snooze is ephemeral and time-boxed, so it does not survive a window reload. The worst case is that auto-import resumes up to five minutes early. That is benign, self-correcting, and one click to redo, against roughly 30 lines and a new class of persisted state with its own failure modes.
+
+This is a decision with history. Snooze once wrote `general.autoImport: false` into global settings and relied on an interval to write it back; a reload mid-snooze left auto-import off permanently. The fix moved the state into memory.
+
+**Revisit when** snooze durations grow long enough that losing one is expensive, or the state stops being purely time-boxed.
+
+The comment at `StatusBarHandler.ts:65-69` currently defends this too broadly. It says persisting snooze "could leave auto imports off forever", which is true of a persisted flag and false of a persisted end-timestamp that expires by wall clock. The comment needs narrowing to the case it actually argues against; that is part of the dead-state cleanup ticket.
+
+## Keep the status bar item for the life of the window
+
+The item appears only after a Verse file activates the extension, so it never shows in windows that never touch Verse. Inside a Verse session it is the snooze surface and the only discoverable route to the menu. Hiding it when the active editor is not Verse costs a show-and-hide state machine, and makes a running snooze countdown disappear while it is still counting.
+
+**Revisit when** the status bar item stops being the primary entry point, or a user reports the footprint as a problem.
+
+## Do not localize yet
+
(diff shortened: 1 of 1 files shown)


vukefn:
## Fix order for the campaign's tickets

From `docs/review/lanes/D-synthesis/`, with its gate's amendments applied. Forty-one tickets are a list, not a plan; this is the plan.

Three edges shape it. **#386 before the writer refactors** — land the safety net, then rebuild behind it. **Decisions that can delete work come before the work** (#379/#380 before #382). **#392 → #395 → #396**, because the case rule's home is the unified resolver.

The lane A cluster (positions 22-29) touches activation, manifest, commands and status bar, which the B/C/G/H chains barely touch. It parallelises with everything above it.

### First five

1. **#372** — digest parser, regeneration, invariant sweep. The only defect shipping wrong data today: 8 phantom entries ride every .vsix, one status-menu click from silent wrong edits under default `autoImport`. The durable fix is the whole-artifact sweep, not the regex — the premise re-expires at every Epic refresh.
2. **#391** — fail closed when the visibility writer cannot read a declaration file. The worst credible failure: the project stops compiling, from a read failure the user never saw, under a success message. Two-line fix that already exists twenty lines away.
3. **#386** — verify the rewrite before applying it. Placed on dependency, not severity. One `scanModuleImports(organized)` comparison converts every composition bug, current and future, from silent text loss into a refused edit.
4. **#373** — one lexer and one declaration-head matcher. The chokepoint: every lexical fact after this lands once instead of four times.
5. **#374** — line comments do not stop a `<#`. Immediately on top of #373 so it lands once. Flip tests 143 and 178 to pin the compiler's answer, citing `ParametricFunctions.versetest:253-256`.

### Tier 2, positions 6-13

**#375** (rides #372's change window) · **#393** · **#385** · **#383** · **#400** · **#390** · **#392** (fold #397's item B in) · **#395** · **#396**

### Tier 3, positions 14-17 — decisions first, then the refactor

**#379** and **#380** are decision tickets. Both can delete work: module-only scanning turns #382's pattern fixes into deletions, and dropping persistence closes five staleness items at once. Then **#384**, the single rebuilder, behind #386 and #390. Then **#389**.

### Tier 4, positions 18-21

**#378** · **#377** · **#376** (the discovery test is the part that must not slip) · **#398**

### Tier 5, positions 22-29 — lane A, parallelisable

**#364** first (one line; converts the engines floor from promise to enforcement) · **#359** · **#360** · **#361** · **#365** · **#363** · **#362** · **#369**

### Tier 6, positions 30-38

**#387** · **#394** · **#366** · **#367** · **#368** · **#381** · **#382** · **#388** · **#397**

### Amendments the gate made to this order

**The #386 dependency is narrower than the synthesis claimed.** It is hard for **#384** (full-document replace). It is half-real for **#389** — the rescan catches the unclosed-comment shape but not the comment-interior insert, which is invisible to the scanner by construction. It is **not** a dependency for **#383** or **#385**, whose defect classes pass an import-set-plus-line-survival check. Those two can start immediately.

**#399 is not absorbed and stays open**, blocked on #392 and #395. The synthesis proposed closing it; the gate reversed that, because `resolveModuleLocations` sorts candidates alphabetically, so no ancestor-or-self rule arrives by construction from either ticket.

**#400 is new**, filed after the gate confirmed G-03 was genuinely uncovered. It is the campaign's only confirmed compile-break-class finding, and #384's fix text never named the rule. Whoever works #384 should either close #400 as absorbed with the acceptance text naming the group-order rule explicitly, or fix it first.

**#370 is closed** as deferred on cost and benefit, not rejected.

### Still waiting on the maintainer

**One live compile.** `X:int = 0, using {/Verse.org/Simulation}` and `Y:int = 0, G := import(/Verse.org/Simulation)` at file scope. It gates **#381**'s B-14 item and nothing else. Every other Verse question the campaign raised was settled from Epic's compiler source.

**Two lane A decisions never filed**, because both change behaviour rather than fix a defect: replacing the hover-sensor CodeLens trigger with selection-driven visibility, and adding a bundler.

vukefn:
Closing: the review record and its decisions are kept locally rather than in the repository. Nothing from the campaign lands here.

The 41 tickets it produced (#359-#400, less the two closed) stand on their own — each carries its evidence, line references and smallest fix in the body.

PR #411, opened 2026-08-15
fix(imports): give Optimize Imports the diagnostic evidence

### LINKED ISSUE #385: Optimize Imports can report success on the failure it was invoked to fix
## What

`buildOrganizedContent` hardwires `NO_DIAGNOSTIC_LINES` on the strength of a comment that is false about the system. Optimize Imports therefore places a diagnosed fix where it fixes nothing, saves the file, and reports success.

## Where

- `src/imports/ImportDocumentEditor.ts:1149-1152` — the comment "No diagnostic reaches here: organizing is not a fix for one", and the hardwired `NO_DIAGNOSTIC_LINES` at 1152 that rests on it.
- `src/commands/CommandsHandler.ts:155-177` — the only caller. Its own doc at 133-134 says "Anything the compiler currently reports as a missing import is added along the way"; the success toast at 177 fires regardless.

## Effect

A pinned dotted consumer with a missing provider: the quick fix and auto-import place the provider above it (ceiling honored) and the diagnostic clears; Optimize Imports — fed the same path extracted from the same diagnostic — places it below the consumer, saves the file, and shows "Imports optimized successfully" while the diagnostic persists. The two user-facing commands write different documents for one fix, and the one that claims success is the one that fixes nothing.

The divergence manifests only for a bare (rank-1) provider. An absolute provider raises no floor (`resolvesAgainstScopeAbove` false), takes `topExtraPaths`, lands in the block above every pinned line, and fixes the diagnostic on both entry points; a rank-2 provider raises a ceiling on neither path, because `couldResolveAgainst` requires the new path's rank to be below 2.

## Evidence

Mechanical leg: the hardwired empty map at 1152 is load-bearing — it is what makes organize refuse the evidence.

Epistemic leg, and the reason this is larger than its row. The wiring predates the comment. `e3c0640` (2026-07-02, "make Optimize Imports deterministic and atomic") wired `optimizeImports` to `extractImportsFromDiagnostics` to `organizeImports(document, missingImportPaths)` (`src/commands/CommandsHandler.ts:157-164`); the contradicting comment arrived in `b789a86` (2026-08-11), five weeks after that caller existed. A load-bearing comment misstates the system's own call graph, written after the caller it denies, in the voice that stops the next reader questioning it — it will keep recruiting future editors to defend the wrong behavior until it is corrected.

## Smallest fix

Thread an optional `DiagnosticLinesByPath` through `organizeImports` into `buildOrganizedContent`'s `hoistFloor` calls. The parameter already exists on `hoistFloor`, so the plumbing is three signatures. Correct or delete the comment at 1149-1151 in the same change.

---
**Severity:** MEDIUM.
**Source:** lane G import-writing review, rows G-11 (Finding 4) and architect's verdict item 3. Full finding in `docs/review/lanes/G-import-writing/02-findings.md`, re-derived in `03-rulings.md`.

vukefn:
Closes #385

## What was wrong

`buildOrganizedContent` passed a hardwired empty diagnostic-lines map to `hoistFloor`, on the strength of a comment saying "No diagnostic reaches here: organizing is not a fix for one".

One does. `CommandsHandler.optimizeImports` has read the missing paths straight out of the compiler's diagnostics since `e3c0640` — five weeks before the comment denying it was written in `b789a86`.

Without that evidence every pinned import raises a floor, so a bare (rank-1) provider added for a pinned dotted consumer was written *below* the statement that needed it. The quick fix and auto-import, handed the same path from the same diagnostic, wrote it *above*. Optimize Imports then saved the file and showed "Imports optimized successfully" while the diagnostic still stood — the two user-facing commands writing different documents for one fix, and the one claiming success being the one that fixed nothing.

## The rule it turns on

A `using` whose own path resolves through another `using`'s module is document-order-dependent; a bare provider must precede the dotted consumer that resolves through it (compiler source: `SemanticScope.cpp:523-543`, where the resolution walk iterates `GetUsingScopes()` with no position comparison, and `SemanticAnalyzer.cpp:6528-6592`).

The change makes no new claim about Verse. `pinnedBounds`/`couldResolveAgainst` already encode this rule; organize simply now obeys the one add already followed.

## What changed

- `ImportSuggestionExtractor.extractImportsFromDiagnostics` returns `{ paths, diagnosticLinesByPath }`. The line each diagnostic was reported on already existed in that loop and was discarded there — it is the last place it exists. Both halves come out of the one pass so they cannot disagree about which diagnostic named which path.
- `DiagnosticLinesByPath` moves to `src/types/moduleInfo.ts` so the extractor and the editor share one definition rather than two shapes that must stay compatible.
- The map threads through `ImportHandler.organizeImports` → `ImportDocumentEditor.organizeImports` → `buildOrganizedContent`, each optional and defaulting to no evidence.
- The false comment is corrected rather than deleted: a diagnostic does reach there, and a pinned import the evidence shows to be the consumer raises a ceiling the top block satisfies rather than a floor.
- `CommandsHandler.optimizeImports` passes both halves on, which makes its own doc true as written.

Only the extras loop changes behaviour. `groundedPaths`/`crossesPinnedProvider` read no diagnostics and answer a different question, and are untouched.

## Tests

Four regression tests, each verified to fail with the fix reverted and pass with it restored:

- `buildOrganizedContent` places the bare provider above the pinned dotted consumer the diagnostic reports on — and its twin, with the diagnostic pointing elsewhere, still leaves the written order alone.
- `organizeImports` carries the evidence across every hop, so a parameter dropped mid-chain is caught rather than only the rebuild being right.
- `CommandsHandler.optimizeImports` hands the organizer the map it extracted.

Two more pin the extractor's new half: lines concatenate where two diagnostics name one path, and a message it declines to import from records none.

Test fixtures that cast `{ message }` to `vscode.Diagnostic` now carry a real `Range`. Reading the line defensively in production instead would fail open and hide a genuinely absent range.

## Review

One round, fresh-context reviewer at opus. Verdict `PASS_WITH_SUGGESTIONS`, no Critical findings. It probed twelve families rather than reading alone: every grouping/sort/syntax permutation, CRLF, indented `using:` pairs, `;`-shared and comment-anchored pins, multiple pinned imports, stale and out-of-range evidence, idempotence, and the full command chain against a real `ImportHandler`.

Two results worth recording:

- **Idempotence holds.** Feeding the organized output back produces byte-identical text and makes zero `applyEdit` calls, so the `organized === text` short-circuit still fires — including when the evidence handed back is stale.
- **The residual is reached in practice.** Where a *second*, undiagnosed pinned import raises a floor, the provider still lands below the consumer and the toast still fires. That placement is correct by `hoistFloor`'s own rule — a compiling file outranks a fixed diagnostic — so this PR narrows the reported class rather than closing it.

The reviewer's one `Important` finding is pre-existing and not introduced here: a diagnostic about a statement sharing a pinned line via `;` can license the override for a pinned import that is the genuine provider. `addImportsToDocument` produces the identical output on the identical input today, and `couldResolveAgainst` already documents the trade-off. Narrowing it needs a column on `ScannedImport`; that belongs in its own ticket.

Suggestions on code this PR introduced were applied in `3bf58e4`: the extractor's parallel Set is gone (the evidence map's keys are already the deduplicated paths, which makes the invariant structural rather than a convention), the statement-keyed map got a name of its own so it cannot be swapped with the path-keyed one, and the new comment and one test name were corrected where they overstated.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> verse-auto-imports@0.11.1 test
> jest --verbose

Test Suites: 44 passed, 44 total
Tests:       1300 passed, 1300 total
Snapshots:   0 total
Time:        8.027 s
Ran all test suites.
```

`npm run format:check`

```
> verse-auto-imports@0.11.1 format:check
> prettier --check "**/*.ts"

Checking formatting...
```

1292 tests before, 1300 after — exactly the eight added, so no existing case silently disappeared.


vukefn:
The review gate's one `Important` finding is filed separately as #412: a diagnostic about a statement sharing a pinned `using` line via `;` can license the consumer override for a pinned import that is the genuine provider. Pre-existing on both entry points and documented at `couldResolveAgainst`; not a regression from this PR, and narrowing it needs a column on `ScannedImport`.

Issue #410, opened 2026-08-15
One declaration-head matcher instead of three


vukefn:
## What

Split out of #373, whose lexical half landed separately. Three files spell "what does a Verse declaration head look like", and a fix to that grammar has repeatedly landed in one of them and not its siblings.

## Where

- `src/services/digestParsing.ts` — a generic head plus a keyword lookup: `DECL_RE` (:65), `PARAM_TYPE_HEAD_RE` (:68), `PARAM_TYPE_TAIL_RE` (:71), `DECL_KEYWORD_RE` (:74), assembled by `matchParametricTypeHead` (:97).
- `src/services/AssetsDigestParser.ts` — `CLASS_OR_STRUCT_DECL` (:18) and `INSTANCE_DECL` (:26).
- `src/services/ProjectPathScanner.ts` — eight patterns at :295-311, plus `DECLARATION_KEYWORD_AFTER_ASSIGN` (:21) reached from the fall-through branch at :610.

## Effect

Divergences #373 named that this half still owns:

- Enum handling differs in all three. `enumPattern` (`ProjectPathScanner.ts:304`) accepts at most one specifier group where the class, struct and interface patterns beside it accept any number, so `Colour<native><public> := enum:` is dropped by the one and kept by the others.
- `digestParsing.ts:270` skips receiver-style extension methods outright, where `ProjectPathScanner`'s `extensionMethodPattern` (:308) indexes them.
- Only `ProjectPathScanner` masks comments before matching; both digest parsers rely on their input being machine-generated, which is documented but is a fourth answer to "is this line a declaration".

And one this repo learned while fixing #373's C-04: a keyword rule inside `ProjectPathScanner` alone has to be spelled twice, once in the per-type pattern and once in the fall-through backstop. C-04 was filed against the patterns only; bounding those without the backstop turned a mistyped declaration into a dropped one, because the backstop was a bare `line.includes("class")`. A single missing boundary therefore had two independent homes inside one of the three matchers.

## Smallest fix

One shared head-matcher module returning `{ name, specifiers, keyword, operator }` from a trimmed line, with per-caller policy on top — the digest index wants module/class/function/variable, the assets parser wants class and struct names, the project scan wants all of them plus visibility. `digestParsing.ts` is already close to this shape and is the natural donor.

One shared probe corpus run through all three callers, as #373's lexical half now has for the lexer, so a future divergence fails loudly rather than waiting to be found by a review.

## Acceptance

[...]

PR #413, opened 2026-08-15
refactor(imports): read Verse comment and literal structure in one lexer

### LINKED ISSUE #373: One lexer and one declaration-head matcher instead of seven grammars
## What

The repo re-spells the Verse grammar seven times: four readers of comment/string/indent lexical state, and three independent "what does a declaration head look like" matchers. Every lexical failure found in the lane B review is two of those four readers disagreeing with each other rather than the extension disagreeing with Verse in a considered way, and every declaration-grammar fix so far has landed in one of the three matchers and not its siblings.

## Where

Four lexical readers:

- `lexLine` (ImportScanner.ts:173-271) vs `maskCommentsAndStrings` (verseText.ts:31-114) — two comment/string lexers.
- `indentWidth` (ImportScanner.ts:273-275, one unit per whitespace character of any kind) vs `indentOf` (verseText.ts:141-153, tab counts as four) — two indent-width models feeding the same `<#>` marker-body rule.
- `bracedUsingSpan`'s inline brace count (ImportScanner.ts:472-491, running on `codeOutsideLiterals`, which leaves `'` literals unmasked, and with no paired-quote skip) vs `countBraces` (verseText.ts:177-179) — two brace counters.

Three declaration-head matchers: `digestParsing.ts`, `AssetsDigestParser.parseDigestContent`, `ProjectPathScanner.extractDeclarations`.

## Effect

Divergence instances the two verdicts cite:

- B-01a — a `<#>` in a marker body leaks `blockDepth` out of `blankCommentRange` (verseText.ts:206-211, carried out at :48) while `classifyLines` does not (ImportScanner.ts:201-203), so the masker and the scanner disagree about which lines are comment.
- B-03 — the masker has no `'` state (verseText.ts:98-107), so `X := '#'` blanks the rest of the line and `X := '"'` opens a string; `lexLine` reads both correctly (ImportScanner.ts:212, 242-243).
- B-05 — four of five probe classes diverge: `'#'`/`'"'`, the suffix `Shop'a#b'`, the interpolation `"a{F("#")}b"`, and the marker-body leak.
- B-07 — the two width models end a marker body on different lines on mixed tab/space files.
- B-18 — `countBraces`' paired-quote skip misses the escape literal `'\{'`, so brace depth drifts file-wide.
- C-04 — `AssetsDigestParser.ts:18` spells `(?:class|struct)\b`; `ProjectPathScanner.ts:283-285` omits the boundary, so `Foo := classify_items(X)` records a class. The repair existed one file away when the bug shipped.
- Enum handling differs in all three matchers; `enumPattern` (ProjectPathScanner.ts:286) also accepts at most one specifier group where the class/struct/interface patterns accept any number.
- `digestParsing.ts:233-236` skips receiver-style extension methods while `ProjectPathScanner.ts:527-548` indexes them.
- The scanner masks comments; both digest parsers rely on generated-input purity.

A declaration or import right of a divergent construct is read by one half of the extension and not the other. The next Verse lexical fact — a new literal form, a grammar tweak in a UEFN release — must be learned in four places or it becomes the next divergence, and the next declaration-grammar fix will again miss two of three matchers.

## Evidence

One grammar underlies all seven: VerseGrammar.h:1903, 1954, 3221 for the lexical half; the code sites above are the implementations of it.

None of the five lexical divergences has a test that would fail while the two implementations disagree: the masker and declaration suites contain no char-literal-`#`, no interpolation, no `<#>`-in-body, no suffix-`'a#b'` and no mixed-indent fixture, and no test runs both lexers on one probe. The scanner's correct half of two divergence classes is tested (tests 37-48, 179-182), so the gap is exact about divergences rather than about the constructs.

## Smallest fix

One line-lexer producing `{kind, depth, code, maskedCode, indent}`, consumed by both `ImportScanner` and the `verseText` clients, so the divergence class stops being representable and a lexical fix lands once instead of three times. One shared head-matcher module (name, specifiers, keyword, operator) with per-caller policy on top, replacing the three declaration grammars. In both cases, one shared probe corpus run through every caller in a test, so a future divergence fails loudly.

Short of the full unification, the per-site interim repairs are: the paired-quote skip applied to `#` and `"` in the masker (the fix behind B-03 and half of B-05); the `'\{'` extension to `countBraces`' skip (B-18); one width function instead of two (B-07); `\b` after the three keywords in `ProjectPathScanner` (C-04).

---
**Severity:** MEDIUM (both verdict items).
**Source:** lane B review, verdict item 2 with row ABS-1 and rows B-01, B-03, B-05, B-07, B-18; lane C review, verdict item 4 with row C-04. Full findings in `docs/review/lanes/B-verse-reading/02-findings.md` and `docs/review/lanes/C-digest-project/02-findings.md`, re-derived in each lane's `03-rulings.md`.

vukefn:
Closes #373

Takes the lexical half of #373: the four separately-implemented readers of Verse comment, string and indent structure become one, consumed by both the import scanner and the `verseText` clients. C-04's missing word boundary comes along, since it is a one-token repair in the same report.

The second half — one declaration-head matcher in place of three — is deferred by agreement and filed as #410.

## What was wrong

Every lexical failure the lane B review found was two of these four disagreeing with each other rather than with Verse:

| | |
|---|---|
| B-01a | a `<#>` in a marker body leaked block-comment depth out of `blankCommentRange`, where `classifyLines` did not |
| B-03 | the masker had no `'` state, so `X := '#'` blanked the rest of the line and `X := '"'` opened a string |
| B-05 | five probe classes diverged, including the interpolation `"a{F("#")}b"`, which the masker read flat |
| B-07 | two indent models ended a marker body on different lines of a mixed tab/space file |
| B-18 | `countBraces`' paired-quote skip missed the escape literal `'\{'`, so brace depth drifted file-wide |
| C-04 | `ProjectPathScanner` omitted `\b` after three keywords, so `Foo := classify_items(X)` recorded a class |

## What changed

**`src/utils/verseLexer.ts` (new).** One line lexer. Given a line and the `{ depth, markerIndent }` the lines above left open, it returns the block-comment depth either side, whether the line is in or opens a `<#>` body, and three views of the text: `codeWithoutComments` (comments deleted so a spliced token rejoins), `codeOutsideLiterals` (literal contents spaced), and `masked` (same length as the input, for offset-preserving callers).

**`src/utils/verseText.ts`.** `maskCommentsAndStrings` becomes a driver over it. `countBraces` **loses its paired-quote heuristic** rather than gaining an `'\{'` case: char literals are blanked upstream, so a rule about which braces are real lives in the lexer where every reader gets it. `scanBraces` is the one brace primitive, reporting the depth after a range and the column where it first returned to zero. `indentOf` moves into the lexer and is re-exported here, so there is one width model with one definition.

**`src/imports/ImportScanner.ts`.** `lexLine`, `scanLine`'s frame types, and `indentWidth` are deleted — 287 lines down to a 4-line driver. `bracedUsingSpan`'s inline brace loop now calls `scanBraces`.

**`src/services/ProjectPathScanner.ts`.** C-04, in both places it lived — see below.

### The rules, and what settles each

Every claim is cited to the compiler's own `VerseGrammar.h`, not read off the readers under repair:

- **A `'` preceded by `[A-Za-z0-9_]` opens a quoted segment suffix; a `'` anywhere else opens a char literal.** `Ident := Alpha {Alnum} !Alnum ["'" … "'"]` (:1954) reaches for the suffix only straight after the alnum run, while `CharLit` is dispatched from prefix position alone (:3497-3501). Both old readers' comments assert this distinction is impossible — `verseText.ts:28-29` and `ImportScanner.ts:261-266`. It is exactly decidable, and that is what lets char literals be masked while `Economy.Shop'Loc'` survives intact.
- **`<#` opens a real block comment at every place; `<#>` advances depth at none.** `Text`'s `case '<'` (:2115-2137) dispatches `BlockCmt()` everywhere including inside a marker body, while `<#>` reaches `IndCmt()` at trivia, code and marker-body place only and is three characters of plain text elsewhere. This is B-01a, and reading the whole clause rather than the one arm fixes two more cases both old readers got wrong *together*: a `<#>` inside a block comment, and a `<#>` inside a string literal — see below.
- **`'\{'` is a real escape literal** — `IsStringBackslashLiteral` (:381) includes `{`.
- **A quoted suffix forbids `{ } " \` and allows a bare `#`** — `IsIdentifierQuotable` (:377). That is why leaving the suffix standing in `masked` costs nothing.
- **Indentation is compared byte-wise against the opener's prefix, and an inconsistent line is error `S89`** (`Line`, :1666-1690). So the two width models agreed on every file that compiles, and unifying on tab-counted-as-four is the whole of B-07.

### One divergence kept, and named

The two readers want **opposite** fallbacks when a line ends inside an open literal, and both are right. The masker must fail towards blanking — text it wrongly offers as code becomes a declaration `moduleDeclarations` edits as if it were real, which `moduleDeclarationsMasking.test.ts:56` pins. The scanner must fail towards keeping — a `using` it misses is read as permission to remove an import, which `ImportScanner.ts` has documented since #272.

So the lexer reports `endedOpen` and each driver chooses, under a doc comment saying why. Accidental divergence stops being representable; this one gets a name.

### A third case the same clause settles

Reading the whole `case '<'` arm turned up one more, found at the review gate: **a `<#>` inside a string literal opened an indented-comment body over every line below it**. `Label := "a<#>b"` followed by an indented `using` returned no paths at all — and an empty answer from `allUsingPaths` is what its caller reads as permission to remove an import.

Pre-existing and identical on `origin/main`, so not a regression, and not a divergence either — both old readers agreed here and were both wrong. It is in scope because the single lexer had cited that clause in its own header and claimed the rule held in every scope while implementing three of its four places. One lexer documenting a rule it three-quarters keeps is worse than four readers that never claimed to agree.

### C-04 had two homes, not one

The ticket cites the class/struct/interface patterns. Adding `\b` there alone made things **worse**: the fall-through backstop was a bare substring test —

```js
if (line.includes(":=") && (line.includes("class") || line.includes("struct") || ...)) continue;
```

— so `Items := classify_items(X)` went from *wrongly recorded as a class* to *dropped entirely*. Both are bounded here, the backstop via a named `DECLARATION_KEYWORD_AFTER_ASSIGN` anchored at the `:=`. This is #373's own thesis in miniature and it is why #410 is worth doing.

## Tests

`src/utils/__tests__/verseLexerCorpus.test.ts` is the shared probe corpus the ticket asks for. Seventeen probes, each run through both entry points, each asserted twice: that the readers **agree**, and that what they agree on is the language's answer. The two teeth are independent — the `<#>`-inside-a-block-comment and `<#>`-inside-a-string probes fail only their correctness half against the old code, because both old readers agreed there and were both wrong.

A further block runs probes through the readers built on those entry points — `allUsingPaths`, `scanModuleImports`, `scanModuleAliases`, `findExplicitModuleDeclarations` and `ProjectPathScanner.extractDeclarations` — since those are where a lexing mistake is finally paid for. One of them pins a contrast worth naming: `allUsingPaths` reads `codeWithoutComments` and so **over**-reports a `using` written inside a string, while `scanModuleImports` reads `codeOutsideLiterals` and reports none. Opposite directions, both deliberate — one decides whether removing an import is safe, the other decides what already counts as imported.

**Proved to detect the defect.** Against `1b6244a`'s readers, 16 of the 37 corpus assertions fail:

```
● a `#` inside a char literal is content › both readers agree on which text is a statement
● a `#` inside a char literal is content › and what they agree on is the language's answer
● a `"` inside a char literal opens no string › (both)
● a `#` inside a string nested in an interpolation is content › (both)
● a `<#>` inside a marker body opens no block comment › (both)
● a `<#>` inside a block comment opens no nested block › and what they agree on is the language's answer
● a `<#` inside a marker body outlives the body › (both)
● a marker body indented with a tab ends where a two-space line meets it › (both)
● a marker body runs through a blank line › (both)
● brace depth › returns to zero across an escaped brace in a char literal
```

and all three C-04 cases fail against the pre-fix `ProjectPathScanner` while its four positive cases still pass. The two `<#>`-in-a-string probes were checked the same way against the lexer before that fix.

That check earned its keep twice. The first version of those two probes **passed** against the defect: the marker sat at column 0 and the line below it was also at column 0, so a wrongly-opened body closed again immediately. Both now indent the lines below.

Also added: `verseText.test.ts` gains direct cases for `maskCommentsAndStrings` and `scanBraces`, which had none — the offset and length contracts `moduleDeclarations` depends on.

No existing test was modified or removed. The diff against `origin/main` is 307 test-file insertions and one deletion, that deletion being an import line.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> verse-auto-imports@0.11.1 test
> jest --verbose

Test Suites: 45 passed, 45 total
Tests:       1354 passed, 1354 total
Snapshots:   0 total
Time:        8.485 s
Ran all test suites.
```

Baseline on `origin/main` was 44 suites / 1292 tests. All 1292 still pass unmodified; the 62 added are the corpus, the `verseText` contract cases and the C-04 regression.

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

`npx jest src/imports/__tests__` — 9 suites, 825 tests, all passing, read in full rather than by summary: a shared lexer that changed a rule surfaces as an unrelated existing case changing shape, not as the failure you predicted.

## Review

Two rounds against a fresh context, at the session model. **Round 1: PASS_WITH_SUGGESTIONS**, no Critical. **Round 2: PASS**, scoped to the delta.

The reviewer verified all eight grammar claims against `VerseGrammar.h` itself rather than taking them, and ran its own differential: on 46 real `.verse` files including the 600 KB `Fortnite.digest.verse`, old and new produce byte-identical output from `maskCommentsAndStrings`, `classifyLines`, `scanModuleImports`, `allUsingPaths` and `scanModuleAliases`. A 200,000-case fuzz over the offset contract found no violation; a 150,000-case fuzz seeded with `<#>` and quote tokens found 976 differences, every one of which contains **both** a marker and a quote — nothing outside the intended fix moved.

Round 1's single `Important` was the `<#>`-in-a-string case, fixed here rather than deferred. Its suggestions taken: one width model with one definition, `LexedLine.indent` dropped as unread, the `scanBraces` doc corrected, incident narration trimmed, and the corpus extended to the derived readers. Round 2 left two comment-accuracy suggestions, both applied — one of them a counterexample to a sentence this PR had added, which is recorded below.

### A pre-existing defect the review surfaced

Not introduced here and not fixed here, but worth stating plainly rather than leaving in a comment: **a `"` interpolation spanning lines leaves `maskCommentsAndStrings` output brace-unbalanced.** Each line is masked from the point a literal opened to the line's end, so `Msg := "Score: {` over `Points}"` loses its `{` and keeps its `}`, and the file-wide depth ends at −1. The fallout is real — on a module whose body contains such a string, `ProjectPathScanner.extractDeclarations` reports the nested declarations at top level, because the stray `}` closed the enclosing body a line early. Byte-identical on `origin/main`.

The `scanBraces` doc now states the limit instead of claiming balance, so the next person to meet a stranded depth is not reassured out of looking.

## Not in this change

- The three declaration-head matchers, beyond C-04 — #410.
- `LineCmt`'s treatment of `#>`. The grammar (:2105-2114) makes a `#>` after a `#` inside a block comment error `S05`, where both readers here close the block. That is over-acceptance on a file that does not compile, so it is neither a divergence nor a defect a user can reach.

### DIFF
--- a/changelog.d/373.fixed.md
+++ b/changelog.d/373.fixed.md
@@ -1,0 +1,1 @@
+**Verse file scanning**: a `using` or declaration written beside a char literal, an interpolated string or a nested comment marker is no longer missed, and a call named after a type keyword is no longer recorded as a type
--- a/src/imports/ImportScanner.ts
+++ b/src/imports/ImportScanner.ts
@@ -1,4 +1,6 @@
 import { ImportFormatter } from "./ImportFormatter";
+import { LexedLine, LexState, lexVerseLine } from "../utils/verseLexer";
+import { scanBraces } from "../utils/verseText";
 
 /**
  * Splits text into lines without keeping the carriage return of a CRLF pair.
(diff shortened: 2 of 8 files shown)


vukefn:
Two issues linked from this branch:

- #410 — the deferred half of #373, one declaration-head matcher in place of three. Split at the maintainer's direction; this PR takes the lexical half plus C-04.
- #414 — a pre-existing defect the review gate demonstrated. A multi-line string interpolation leaves `maskCommentsAndStrings` output brace-unbalanced, so `ProjectPathScanner` reports nested declarations at top level. Byte-identical before and after this PR, so it is neither introduced nor fixed here; `scanBraces`' doc now states the limit rather than claiming balance.

PR #418, opened 2026-08-15
fix(imports): read a block comment opened inside a line comment

### LINKED ISSUE #374: Line comments do not stop a `<#`: both scanners read live code the compiler treats as comment
## What

Both scanners hold a documented, test-pinned belief that a `#` line comment runs to end of line, so a `<#` inside one is text rather than an opener. Epic's compiler and Epic's own shipped TextMate grammar both say otherwise: a `<#` inside a line comment opens a real, multi-line block comment. Separately, a `<#>` written inside a string, a line comment or a block comment is plain text to the compiler, and both lexers read it there as a marker or as a `<#` opener.

## Where

- ImportScanner.ts:134-135 — the clause "so a `<#` inside one is text rather than an opener" (commit e76553e, #163, with no evidence the compiler was consulted). verseText.ts:23-25 states the end-of-line rule without that clause. The pins are scanner test 143 and classifyLines test 178 only; the masking suite pins nothing here.
- verseText.ts:206-211 — `blankCommentRange` matches `content.startsWith("<#", i)` on the `<#` of a `<#>` in a marker body, increments `blockDepth`, and the depth is carried out of the body at :48. `classifyLines` does not leak on the same text (the `<#>` return at ImportScanner.ts:201-203 leaves `nesting` unchanged, the marker update at :353 is gated), so the two lexers also diverge with each other.
- ImportScanner.ts:189-191 and verseText.ts:58-64 — the `nesting > 0` / `blockDepth > 0` branches count the `<#` of a `<#>` as a nested opener.
- ImportScanner.ts:201 before :205, and verseText.ts:70 before :84 — the `<#>` test runs before the literal state, so `"a<#>b"` reads as a marker.
- ImportScanner.ts:624-634 — `anchorsCommentBelow` re-lexes the span with `scanLine` and inherits the same digraph defect.

## Effect

In the region the compiler reads as comment, the extension reads live imports: presence that suppresses a needed `using` (the user sees an unknown-identifier error the quick fix then refuses to fix), and organize rebuilds lines inside text the compiler considers dead. This reaches any user who writes an unclosed `<#` inside a prose comment — rare, but not exotic in comments about comments.

The three `<#>` interior cases:

- A `<#>` inside a `<#>` body leaks `blockDepth` 1 past the body's end, masking everything to the next stray `#>` — usually the rest of the file. `findExplicitModuleDeclarations` then misses every later declaration and the visibility writer adds a second part with a different specifier, breaking the compile; `ImportPathConverter` and `ProjectPathScanner` go equally blind.
- A `<#>` inside a `<# ... #>` block bumps nesting in both lexers, so the real `#>` no longer closes the block and every import below reads as comment: presence misses, duplicates written.
- A `<#>` inside a string ends the line's code at the marker and marks the indented lines below as comment; a `using` there stops counting and `allUsingPaths`' empty answer becomes permission to remove an import the file needs.

The structural half is worse than any single trigger: `classifyLines`' `depth` and the masker's `blockDepth` are file-long accumulators with no resynchronization and no sanity bound, so one wrong increment converts the entire remainder of a file into comment and nothing downstream notices — the scans just return less. To the user this presents as the extension "randomly" ignoring half a file, which is close to undebuggable from a bug report.

`anchorsCommentBelow` over-flags only: a closed trailing block comment containing a `<#>`, such as `using { /A } <# see <#> marker #>`, freezes a rewritable import in place.

## Evidence

- `LineCmt := '#' !'>' {Text} Ending` (VerseGrammar.h:2016). At LineCmt place the jump table stops on `<#` (`S[1] != '#' ? 1 : 0`, VerseGrammar.h:462-464), the `case '<'` branch at 2117-2126 calls `BlockCmt()` with no place guard, and `Text<BlockCmt>` consumes newlines (2099-2103).
- Epic's shipped TextMate grammar encodes the same rule: `LineCmt` (`begin (?<!<)#(?!>)`, `end $`) includes `#BlockCmt`, whose multi-line region suppresses the parent's end (`book-of-verse/docs/Assets/Verse.json:12-27`).
- Compile-validated: `book-of-verse Tests/ParametricFunctions.versetest:253-256` holds a `<# ... #>` inside a `#` line comment. Under the repo's "text" reading its `#>` would be a parse error — a bare `#>` inside a line comment is S05 (the jump table for `#` stops on `#>`; the `case '#'` else-branch at 2113-2116), and both digraphs are excluded from `Printable` (412) — so the file compiles only under the nested-block reading. It is the one compile-validated artifact in the book that discriminates between the two readings. No versetest exercises the multi-line case and none contradicts it.
- For `<#>`: at VerseGrammar.h:2122-2139 the `<#>` case parses a nested `IndCmt` only when `ParsePlace` is `Space`, `Content` or `IndCmt` (the constexpr guard at 2128); at `BlockCmt`, `LineCmt` and `String` place it falls to `Next(3)` of plain text (2136-2138). `Comments.versetest:308-336` corroborates nested `<#>` as nested indented comments at `IndCmt` place.

## Smallest fix

- In `lexLine`, on the `#` return path, first scan the comment tail for an unmatched `<#` and carry its depth; mirror it in the masker.
- Give both lexers an explicit literal-interior `<#>`-as-text branch that consumes all three characters. Moving the `<#>` test after the literal-state check, or gating it on `innermost === undefined`, is not enough on its own: the `<#` test deliberately still runs inside literals, because a `<#` in a string is a real opener (tests 40 and 276), so a bare move hands the first two characters of a string-interior `<#>` to the `<#` branch and converts the text case into a false opener.
- Flip scanner test 143 and classifyLines test 178 to pin the compiler's answer, and use a fixture that contains a later `#>`. With no `#>` following anywhere the compiler backtracks out of `BlockCmt` and raises a hard S04 parse error (VerseGrammar.h:2036-2037) rather than silently commenting the file, so test 143's current two-line fixture does not exhibit the divergence at all; a subsequent block comment's closer suffices.
- Masking a nested `<#>` as text at `IndCmt` place stays behaviourally equivalent to the compiler, because a nested indented-comment body is always a subset of the outer body's line range.
- `anchorsCommentBelow` needs nothing beyond the digraph fix.
- Worth considering with it: log loudly in debug output when a file ends at depth > 0 with no real unclosed opener, or when the masker and the scanner disagree on a line's kind.

---
**Severity:** MEDIUM.
**Source:** lane B review, verdict items 1 and 3, rows B-01 and B-27. Full finding in `docs/review/lanes/B-verse-reading/02-findings.md`, re-derived in `03-rulings.md`.

vukefn:
Closes #374

`LineCmt` runs over the same `Text` as every other place, and `Text`'s `case '<'` dispatches `BlockCmt()` with no place guard (`VerseGrammar.h:2115-2126`), so a `<#` written in a `#` line comment opens a block comment that runs to its `#>` however many lines below. The lexer consumed the comment tail character by character instead, so every line the compiler reads as comment was read here as live code.

## Scope

Most of #374 landed two hours earlier in #413, which replaced the four separate readers with `src/utils/verseLexer.ts`. That took the ticket's three `<#>`-interior cases and the scanner/masker divergence with it. What was left is the half named in the title, and it now lives in one place rather than two.

`anchorsCommentBelow` needed nothing, as the ticket predicted: it re-drives the same lexer over the statement's span, so it inherits the rule.

## What changed

**`src/utils/verseLexer.ts`.** The `#` arm no longer consumes the rest of the line. It takes the `#` itself, sets `insideLineComment`, and lets the existing comment-text branch lex the tail — the same branch a block comment's body and a `<#>` marker's body already go through. That branch tests `<#>` before `<#`, so the three rules fall out of one place:

- a `<#` in the tail opens a block, which outlives the line that opened it
- a `<#>` in the tail stays three characters of text, opening neither a block nor a body
- the opening `#` is consumed before the branch is entered, so a `#>` at the head of a line cannot also read as a closer

Every output is byte-for-byte what it was for a comment tail holding neither digraph, `masked` included, so nothing that indexes into it moves.

## The rules, and what settles each

Read from the compiler source at `Engine/Source/Runtime/VerseCompiler/Public/uLang/Parser/VerseGrammar.h`, not from the readers under repair:

- **A `<#` opens a real block comment at line-comment place.** `LineCmt := '#' !'>' {Text} Ending` (:2014-2024) runs the same `Text`, whose `case '<'` reaches `BlockCmt()` at every place — the `if constexpr` guard on that arm covers only the `<#>` case (:2115-2126).
- **The block spans lines.** `Text<EPlace::BlockCmt>` continues over a newline (:2099-2103).
- **A `<#>` at line-comment place is plain text.** The guard admits `Space`, `Content` and `IndCmt`; everything else falls to `Next(3)` (:2127-2138).
- **A bare `#>` at line-comment place is error S05** (:2105-2113), so nothing legal closes at depth 0 inside a line comment.
- **An unclosed `<#` is a hard S04 at EOF** (`BlockCmt`, :2034-2039) — not comment-to-EOF. A file that writes one does not compile; reading its remainder as comment is what the lexer already did for an unclosed `<#` at code scope, so the two agree.
- **Compile-validated corroboration.** `book-of-verse/Tests/ParametricFunctions.versetest:256` writes `<# type{:int|:string}, i.e. any #>` inside a `#` line comment. Under the old reading its `#>` would be S05, so the file compiles only under the block reading. It is the one artifact in the book that discriminates between the two.

## Tests

Three pins encoded the old belief and now pin the compiler's answer. Each fixture gained a later `#>`, because a fixture without one is a file that fails to parse and so does not exhibit the divergence at all:

- `scanModuleImports` › flags an import whose trailing line comment opens a block comment
- `scanModuleImports` › treats a `<#` inside a line comment as the block opener it is
- `classifyLines` › reads a block opener inside a line comment as one

Three probes were added to the shared lexer corpus, where every entry point is held to the same answer: a `<#` in a line comment outliving the line, a line comment that closes its own block still ending with its line, and a `<#>` in a line comment opening nothing over the indented line below it. Two more `classifyLines` cases guard the same-line and ordering halves.

The four assertions that carry the fix were run against the unfixed lexer and fail there.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
Test Suites: 46 passed, 46 total
Tests:       1382 passed, 1382 total
Snapshots:   0 total
Time:        11.697 s
Ran all test suites.
```

`npm run format:check`

```
Checking formatting...
```

## Review

One round, fresh context, `PASS_WITH_SUGGESTIONS`, no Critical. The reviewer re-derived R1-R5 from `VerseGrammar.h` rather than from the diff, ran ~40 probes across `lexVerseLine`, `classifyLines`, `scanModuleImports`, `allUsingPaths`, `maskCommentsAndStrings`, `findExplicitModuleDeclarations` and `ProjectPathScanner.extractDeclarations` — same-line close, close many lines below, never closing, `<#>` with an indented line under it, a line-opening `#>`, a `#>` later in the tail, `#` in a string, a char literal, a quoted suffix and an interpolation, a line comment inside a block comment and inside a marker body, nested `<# <# #> #>`, CRLF and tabs — and asserted `masked.length === line.length` on every one, with no violations. It also compiled the pre-fix lexer standalone to confirm the new fixtures fail against it, and confirmed the `<#>`-before-`<#` ordering is load-bearing by removing it (the probe's answer collapses from `["L1","L2"]` to `[]`).

Two Suggestions applied: a misattributed grammar place in the `#` arm's comment, and an ambiguous "above" in a corpus probe comment.

Three Important findings are left unfixed, all outside this ticket rather than defects in it:

1. **The `<#>` marker's own line has the same gap.** `IndCmt()` lexes the marker's tail with `Text<EPlace::LineCmt>` (`VerseGrammar.h:2041-2047`) — the identical place fixed here — so `<#> note <# opener` should open a block and does not. Same rule, sibling path.
2. **A line comment does not resume after its block closes on a later line.** `Text<LineCmt>` continues after `BlockCmt()` returns, so in `#> using { /X }` the `using` is still comment text; it currently reads as live. Pre-existing, not a regression.
3. **`ImportFormatter.commentStartIndex` truncates a path at a `#` inside a quoted segment suffix.** `IsIdentifierQuotable` excludes `#` only before `>`, so `Mod'a#b'` is legal; the entry is marked rewritable, so Organize Imports can rebuild the line from the truncated path. Unrelated to comments-in-lexing, and the higher-impact of the three.

The ticket's structural note — that block-comment depth is an unbounded file-long accumulator — is unaddressed by design: its own prescribed fix carries no bound.

### DIFF
--- a/changelog.d/374.fixed.md
+++ b/changelog.d/374.fixed.md
@@ -1,0 +1,1 @@
+**Verse file scanning**: a block comment opened with `<#` inside a `#` line comment now reads as commented out to its `#>`, so imports written in that region no longer count as present or get rewritten.
--- a/src/imports/__tests__/ImportScanner.test.ts
+++ b/src/imports/__tests__/ImportScanner.test.ts
@@ -954,10 +954,14 @@ describe("scanModuleImports", () => {
         ]);
     });
 
-    it("does not flag an import whose trailing line comment mentions <#", () => {
-        expect(scanModuleImports(["using { /A } # opens with <#", "using { /B }"])).toEqual([
-            { path: "/A", startLine: 0, endLine: 0, anchorsCommentBelow: false, rebuildLosesText: false, trailingComment: "# opens with <#" },
-            { path: "/B", startLine: 1, endLine: 1, anchorsCommentBelow: false, rebuildLosesText: false, trailingComment: "" },
+    it("flags an import whose trailing line comment opens a block comment", () => {
+        // The line comment ends with its line and the block it opens does not,
+        // so the lines below really are commented out and moving the statement
+        // moves which ones.
+        const lines = ["using { /A } # opens with <#", "using { /Old/Path }", "#>", "using { /B }"];
+        expect(scanModuleImports(lines)).toEqual([
+            { path: "/A", startLine: 0, endLine: 0, anchorsCommentBelow: true, rebuildLosesText: false, trailingComment: "# opens with <#" },
+            { path: "/B", startLine: 3, endLine: 3, anchorsCommentBelow: false, rebuildLosesText: false, trailingComment: "" },
         ]);
     });
 
@@ -1036,9 +1040,12 @@ describe("scanModuleImports", () => {
         expect(scanModuleImports(lines)).toEqual([{ path: "/Verse.org/Simulation", startLine: 2, endLine: 2, anchorsCommentBelow: false, rebuildLosesText: false, trailingComment: "" }]);
     });
 
-    it("does not treat a <# inside a line comment as a block opener", () => {
-        const lines = ["# a block comment opens with <#", "using { /Verse.org/Simulation }"];
-        expect(scanModuleImports(lines)).toEqual([{ path: "/Verse.org/Simulation", startLine: 1, endLine: 1, anchorsCommentBelow: false, rebuildLosesText: false, trailingComment: "" }]);
+    it("treats a <# inside a line comment as the block opener it is", () => {
+        // A line comment is lexed like any other text, so the block runs from
+        // the `<#` to the `#>` below and the import between them is commented
+        // out. Reading it as live is how a needed using gets suppressed.
+        const lines = ["# a block comment opens with <#", "using { /Old/Path }", "#>", "using { /Verse.org/Simulation }"];
+        expect(scanModuleImports(lines)).toEqual([{ path: "/Verse.org/Simulation", startLine: 3, endLine: 3, anchorsCommentBelow: false, rebuildLosesText: false, trailingComment: "" }]);
     });
 
     it("returns entries in document order with correct spans across mixed styles", () => {
@@ -1247,8 +1254,12 @@ describe("classifyLines", () => {
         expect(kinds(["<#> the marker is not a block opener", "code()"])).toEqual(["comment", "code"]);
     });
 
-    it("does not mistake a block opener inside a line comment for one", () => {
-        expect(kinds(["# see <# below", "code()"])).toEqual(["comment", "code"]);
+    it("reads a block opener inside a line comment as one", () => {
+        expect(kinds(["# see <# below", "code()", "#>", "code()"])).toEqual(["comment", "comment", "comment", "code"]);
+    });
+
+    it("ends a line comment that opened and closed a block on its own line", () => {
+        expect(kinds(["# see <# aside #> below", "code()"])).toEqual(["comment", "code"]);
     });
 
     it("keeps a # inside a string literal in the line's code", () => {
--- a/src/utils/__tests__/verseLexerCorpus.test.ts
+++ b/src/utils/__tests__/verseLexerCorpus.test.ts
@@ -137,6 +137,31 @@ const probes: Probe[] = [
         live: ["L1"],
     },
     {
+        // Both readers held the opposite, that a line comment made its tail
+        // text: `Text`'s `<` arm dispatches BlockCmt at every place, so the
+        // block outlives the line that opened it exactly as one opened in code
+        // does. The lines it swallows were being read as live imports.
+        name: "a `<#` inside a line comment outlives the line",
+        source: "# D1 <# D2\nD3 := 1 #>\nL1 := 1\n",
+        live: ["L1"],
+    },
+    {
+        // The line comment ends at its own newline where the block it opened
+        // closed on the line, and the text between the two is trivia either way.
+        name: "a line comment that closes its block still ends with its line",
+        source: "# D1 <# D2 #> D3\nL1 := 1\n",
+        live: ["L1"],
+    },
+    {
+        // The `<#>` arm falls to three characters of text at line-comment place,
+        // so this opens neither a block nor a body. It has to be tested before
+        // the `<#` above or the tail reads as an opener that never closes, and
+        // the indented line below is what tells the two mistakes apart.
+        name: "a `<#>` inside a line comment opens nothing over the lines below it",
+        source: "# D1 <#> D2\n    L1 := 1\nL2 := 2\n",
+        live: ["L1", "L2"],
+    },
+    {
         // Two indent models ended the body on different lines: a tab counted as
         // one column in the scanner and four in the masker.
         name: "a marker body indented with a tab ends where a two-space line meets it",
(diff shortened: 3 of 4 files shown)


vukefn:
The three Important review findings are now tracked, none of them in this ticket's scope: #419 for the other two halves of the line-comment-place rule (a `<#` in a `<#>` marker's own tail, and a line comment not resuming after its block closes), and #420 for `ImportFormatter` truncating a path at a `#` inside a quoted segment suffix.

Issue #396, opened 2026-08-14
Path arithmetic: one separator and one case rule


vukefn:
## What

The path arithmetic has no single separator rule and no single case rule. Byte-exact comparisons, a case-insensitive stat fallback, raw-suffix matching and unvalidated segments coexist across the same walks, so components disagree about when two spellings are one path. Two adjacent assumptions in the same band are equally unguarded: nothing validates an emitted segment against the path grammar, and the scanned universe is pinned to `workspaceFolders[0]` rather than the folder owning the project.

## Where

- **H-08, the `Content` boundary.** Byte-exact: `ImportPathConverter.ts:261-262, 270, 337, 589, 594`; `ModuleVisibilityWriter.ts:165, 239, 308-313`. Case-insensitive stat fallback: `ModuleVisibilityWriter.ts:169-175`.
- **H-09, `path.relative`.** `ImportPathConverter.ts:258, 382, 436, 583`; `ModuleVisibilityWriter.ts:302`.
- **H-10, raw suffix.** `ImportPathConverter.ts:397-403`.
- **H-11, no lexical validation.** Nothing between `searchImplicitModules` / `searchProjectFolderModules` / `searchExplicitModuleDefinitions` output and `buildFullVersePath` (`ImportPathConverter.ts:104-106`).
- **H-12, two URI keys for one disk file.** `ModuleVisibilityWriter.ts:100` (`definitionsUri.toString()`, joinPath-derived from the configured name) against `252` (`result.file.uri.toString()`, findFiles-derived from the on-disk name); the overlap fold at `121-122`.
- **H-14, the wrong workspace folder.** `ImportPathConverter.ts:356, 361, 431-432, 436, 501`; `ModuleVisibilityWriter.ts:70`.

## Effect

**H-08.** With an on-disk `content` under the workspace root, `findContentRoot` succeeds — the definitions file is addressed under a `Content` spelling that exists case-insensitively — while `contentRelativeDirectory` and the converter's scope derivation reject every file, because `startsWith("Content/")` fails against the real casing from `path.relative`. The writer believes the project has no declarations and writes a fresh block; the converter silently loses its scope and near-file search. No message either way. For a workspace opened at `.../content` itself, both sides fail instead — the writer loudly, the converter silently. Either way there is no one rule.

[...]

Issue #395, opened 2026-08-14
Phase 2: delete the bespoke location matching and surface the 100-file cap


vukefn:
## What

Converter phase 2 does its own location matching, which fabricates locations and cannot see nested modules, while the correct implementation — segment-aligned, chain-aware, unit-tested — already exists ten lines away in `moduleLocationLookup.resolveModuleLocations` and is what the cache path uses. Phase 2 also stops at 100 files, with nothing anywhere that distinguishes a completed search from a truncated one.

## Where

- `ImportPathConverter.ts:383-408` — the directory-suffix location model; `397-403` — the parent check, which runs only for multi-segment references.
- `ImportPathConverter.ts:361` — the 100-file cap; `43, 432` — the 5000 cap.
- `moduleLocationLookup.ts:96-168` — `resolveModuleLocations`, the correct matching; `143-156` and `204-218` (`resolveFolderModuleLocations`) compare whole segments.
- `ModuleVisibilityWriter.ts:218-227` — the writer's own scan, deliberately uncapped, with the reason stated.
- The guard table (recon section 8) contains no truncation condition — verified, none exists.

## Effect

**Nested explicit modules (H-06).** A file at `Content/X` declaring `Outer := module:` containing `Inner := module:` has the true path `/proj/X/Outer/Inner`. A reference `Outer.Inner` is skipped, because directory `X` does not end with `Outer` — "Could not find module" for a module the file imports today. A bare reference `Inner` matches the file (the regex sees `Inner := module`), and phase 2 emits location `/X`, building `/proj/X/Inner` — a fabricated path presented unambiguously: unresolvable at best, a namesake's at worst.

**The cap (H-05).** In a project past 100 `.verse` files — routine for team-scale UEFN maps — phase 2 resolves from whichever files `findFiles` returned. A namesake inside the first 100 is written unambiguously, with `isAmbiguous: false` (silent wrong module); a true declaration beyond the cap falls through to phase 3, which cannot see explicit declarations, ending in "Could not find module" for a module that exists.

## Evidence

- Nested explicit modules are ordinary path scopes: `SemanticAnalyzer.cpp:4416-4420` allows module definitions at module scope, and resolution at `6243` walks them.
- `findFiles(..., 100)` result order is unspecified. The compiler resolves over the whole program and treats a multi-match as a loud error (`SemanticAnalyzer.cpp:6295-6312`) — it never silently picks from a sample.
[...]

PR #427, opened 2026-08-15
fix(imports): route both conversion directions through one resolver

### LINKED ISSUE #392: Route both conversion directions through one resolver
## What

The two conversion directions do not share a resolver. Going-absolute runs a three-phase location search; going-relative is pure string arithmetic that never consults the model the class already owns. Nothing establishes that an emitted relative reference resolves back to the module the absolute path named.

## Where

- `ImportPathConverter.ts:649-661` (the whole going-relative conversion) against `473-539` (the going-absolute search).
- `ImportPathConverter.ts:623-680` — `convertFromFullPath`; `findModuleLocations` is never called in it.
- `ImportPathConverter.ts:555-566` — `relativizeAgainst`; `556-561` — the ASCII case fold applied per segment of the whole walk.
- `ImportPathConverter.ts:660-661` — the `belowProjectRoot` gate, bypassed whenever `shortened` is non-empty.
- `ImportPathConverter.ts:614-617` — the doc comment recording the foreign-path pass-through as intended.
- `ModuleVisibilityPlanner.ts:42-47, 68` — the byte-exact counterpart of the same prefix walk.

## Effect

**Namesake capture (H-01).** A namesake module between the importing file and the shared ancestor — a folder `Weapons` beside the file when the import named `/proj/Weapons` — captures the emitted first segment during the compiler's innermost-out walk. The converted file still compiles and imports a different module. Reached by "Use relative path" or "Use relative paths for all" in any project holding a nearer namesake; a same-step clash errors loudly, a nearer single candidate wins silently.

**Foreign-path pass-through (H-02).** A path not under the project is converted instead of refused. A foreign-account path emits `using. other@fortnite.com.Proj.Mod`, whose first segment cannot lex; a foreign-project path under the same account emits a reference that resolves nowhere, or to a local namesake folder. One broken import becomes a differently-broken one that can no longer round-trip back, and a document-wide run writes several at once.

**Fold beyond the project prefix (H-03).** A file under folder `shop` with the import `/proj/Shop/Item`: the fold treats `Shop` and `shop` as one segment and emits `Item`, which resolves to nothing (loud error at the line) or to `shop`'s own `Item` if one exists (silent wrong module). The visibility planner compares the same prefix byte-exact, so the two components disagree about when two spellings are one module — one drift, two treatments, no stated owner.

## Evidence

- No lookup runs on the going-relative direction — verified by reading `convertFromFullPath` whole. Every shortening test uses a namesake-free tree (recon 11.36-11.50).
- `SemanticAnalyzer.cpp:6561-6588` — a non-path `using` argument is an ordinary expression, resolved innermost-out, so a nearer namesake takes the emitted first segment.
- `SemanticAnalyzer.cpp:6243` — path resolution is byte-exact, so two Content-level modules differing only in case can legally coexist. `6243-6293` — a foreign account or sibling project is a loud `ErrSemantic_InvalidScopePath` before any conversion runs, so the H-02 source import could not have compiled; the conversion turns one broken import into a differently-broken one.
- `VerseGrammar.h:1954-1960` — `@` is not an Ident character, so the emitted first segment cannot lex as a reference.
- Epic's own routine would pass a foreign path through too (`SemanticAnalyzer.cpp:12145-12184`), and Epic never lets that happen because the call sits behind the same-package gate at `12234-12236`. The commit that shaped this code, `78fa22fa` ("convert to relative paths the way the compiler does"), therefore argues for the fix: the way the compiler does it includes the input gate. Blame shows no reason recorded for the pass-through, in the doc comment or the commit.
- The repo already documents an instance of the same class: a `using` rooted at a sibling-file alias converts by namesake hunt (`ImportScanner.ts:1322-1329`, recorded as a deliberate boundary in commit `935f6e5`). The acknowledged boundary and H-01 share one missing defense.
- ABS-1 confirmed: no resolve-back exists in either direction; the round-trip lives only in tests, over five friendly trees (recon 11.50).
- The fold's legitimate use case is prefix drift between `.uefnproject` and diagnostic spellings (recon 11.39, 11.42), which only ever needs the first `project.length` segments folded. Epic's routine folds everything (`SemanticAnalyzer.cpp:12157`) but is used only for same-package suggestion text.
- Mitigation, noted for honesty: Epic's own quick-fix suggestions use the same unverified shortening (`SemanticAnalyzer.cpp:12226-12248`), but only for same-package definitions it just resolved. Convert All industrializes the exposure in a way a pasted suggestion does not.

## Smallest fix

ABS-1's mechanism is the smallest installment: after building the relative form, run `findModuleLocations` on it from the document's URI and require the unique answer to rebuild the original absolute path; refuse with a log line otherwise. One extra search per conversion.

Alongside it, in `convertFromFullPath`, require `relativizeAgainst(fullPath, projectVersePath)` to have consumed the whole project prefix — refuse when the path is not under `projectVersePath` — before emitting anything, and retire the doc paragraph at `614-617` with it. And fold only the first `splitPath(projectVersePath).length` segments of the comparison, comparing byte-exact beyond them, at both `relativizeAgainst` call sites.

The target shape is one `resolve(file, reference) -> absolute path` used by both directions, by the presence check, and as the post-write verifier. Without it, every future path feature — aliases across files, qualified segments — re-litigates the same class. The other two facets verdict item 1 names, H-05 and H-06, are filed as the phase-2 ticket.

---
**Severity:** HIGH (H-01, and architect's verdict item 1); MEDIUM (H-02, H-03, ABS-1 — the absence guards HIGH classes).
**Source:** lane H conversion-and-visibility review, rows H-01, H-02, H-03, ABS-1, marker 10.5, architect's verdict item 1. Full finding in `docs/review/lanes/H-conversion-visibility/02-findings.md`, re-derived in `03-rulings.md`.

vukefn:
Closes #392

Going relative was pure string arithmetic that never consulted the module search the other direction runs, so nothing established that the reference it emitted named the module the absolute path named.

## What changed

**The reference is resolved back before it is written.** After building the relative form, `convertFromFullPath` resolves its **first segment** through `findModuleLocations` and requires the one location it returns to rebuild the path that segment must name. First segment, not the whole chain: the compiler resolves only that through the scope chain and then descends byte-exact from whichever match it takes (`SemanticScope.cpp:521-535`), so a nearer namesake carrying nothing below it breaks a reference whose full chain matches elsewhere — and matching the chain as one path would confirm a spelling the compiler cannot read. Anything but one location refuses: two or more mean no relative spelling reaches past the namesake at all.

**A path the project does not contain is refused** rather than shortened by whatever prefix it happened to share. That emitted a first segment containing `@`, which cannot lex (`VerseGrammar.h:1953-1954`), or one naming whatever local module answered to it.

**The ASCII case fold is bounded to the project prefix.** Below it a segment is a module name the compiler resolves byte-exact (`SemanticAnalyzer.cpp:6240-6243`), so folding made `shop` and `Shop` one module. The prefix stays folded because one package legitimately reaches this code spelled two ways. Epic folds the whole path, but only behind a same-package gate (`SemanticAnalyzer.cpp:12145-12184`, `12234-12236`); this adds the gate.

**An importing file the Content root does not hold is refused.** Standing the project root in for the scope was a guess the verification could not catch, because only the search's first phase walks outward from the document. `documentUri` is now required rather than optional, which is where the fallback came from; every caller already passed one.

## Behaviour this removes

Fail-closed, agreed with the maintainer before implementing. A conversion that cannot be confirmed is declined, where before it was written:

- A `.verse` file outside the project's Content root no longer offers the conversion. Two tests that pinned the old root-relative fallback now assert the refusal.
- An explicitly declared target module in a project past `searchExplicitModuleDefinitions`' documented 100-file cap resolves to nothing, so it is refused where string arithmetic used to produce an answer.

## Known limits, all inherited from the search rather than added here

Named in `referenceResolvesTo`'s doc comment so the next reader does not over-trust the check:

- A namesake existing only as an explicit `X := module` declaration is invisible while a folder answers first — the declaration scan runs only when the folder phases return nothing. This is half of H-01, and it overlaps #395.
- On a case-insensitive filesystem the search confirms the casing it asked about, not the casing on disk, so H-03's guarantee is not verifiable through this search on Windows.
- The scanning phases read `workspaceFolders[0]` rather than the folder holding the document, so a multi-root workspace can confirm from a tree the file is nowhere in. That is #396's H-14.

Two things found during review that belong elsewhere, not fixed here:

- `convertAllToRelativePath` reports "No absolute path imports found to convert." when every import was refused, and under-counts on a partial refusal. Filing separately.
- The `<project>/Plugins/<Name>/Content` layout that `ProjectPathHandler` documents supporting is now refused when the workspace is opened at the project root, since `enclosingModulePath` only accepts `Content` at the top of the workspace folder. That is the `Content`-boundary work in #396.

## Review

Two rounds, fresh context each, at `models.review`.

Round 1 returned **FAIL** on a real defect I verified independently before fixing: the fail-closed behaviour was not implemented for an unplaceable file, because `findModuleLocations`' phases 2 and 3 ignore `documentUri` and glob the workspace — so the first segment still resolved uniquely from the project-wide scan. The two tests pinning that refusal were passing only because the suite never stubs `vscode.workspace.findFiles`, which the mock defaults to `[]`. Fixed in 9f21e41, with a regression test that stubs `findFiles` the way production populates it.

Round 2 returned **PASS_WITH_SUGGESTIONS**, having re-run the branch's tests against `origin/main` (8 failures — all three ticket findings genuinely pinned) and against the pre-fix commit (the new guard test fails there for the reason it claims). Its remaining `Important` is the multi-root case above, which `convertToFullPath` already shared and which #396 owns; its suggestions on comment honesty and the `documentUri` signature are applied in 5214b08. Its suggestion to memoize first-segment lookups across a document-wide run is left for the follow-up.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
Test Suites: 46 passed, 46 total
Tests:       1416 passed, 1416 total
Snapshots:   0 total
Time:        12.729 s

$ npm run format:check
Checking formatting...
```

`npx jest src/imports/__tests__` also run whole and read in full — 10 suites, 847 tests, nothing shadowed a neighbouring case.

### DIFF
--- a/changelog.d/392.fixed.md
+++ b/changelog.d/392.fixed.md
@@ -1,0 +1,1 @@
+**Relative path conversion**: converting an import to its relative form now declines when the shorter spelling would reach a different module or none, instead of writing an import that silently points elsewhere.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -657,15 +657,19 @@ export class ImportPathConverter {
      * find. None means nothing here can place the module, which is exactly the
      * position this check exists to stop a conversion being written from.
      *
-     * Two blind spots, both inherited from the search rather than added here.
-     * A namesake that exists only as an explicit `X := module` declaration is
-     * invisible while a folder answers first, because the declaration scan runs
-     * only when the folder phases found nothing. And a folder whose name
-     * differs from the reference only in case satisfies the search on a
-     * case-insensitive filesystem, so this confirms the casing it asked about
-     * rather than the casing on disk.
+     * The check is only as good as the search under it, and that search has
+     * blind spots this does not close - these among them, rather than these
+     * alone. A namesake that exists only as an explicit `X := module`
+     * declaration is invisible while a folder answers first, because the
+     * declaration scan runs only when the folder phases found nothing. A folder
+     * whose name differs from the reference only in case satisfies the search
+     * on a case-insensitive filesystem, so this confirms the casing it asked
+     * about rather than the casing on disk. And the scanning phases read
+     * `workspaceFolders[0]` rather than the folder holding the document, so in
+     * a multi-root workspace they can answer from a tree the importing file is
+     * nowhere in.
      */
-    private async referenceResolvesTo(reference: string, fullPath: string, documentUri: vscode.Uri | undefined, projectVersePath: string): Promise<boolean> {
+    private async referenceResolvesTo(reference: string, fullPath: string, documentUri: vscode.Uri, projectVersePath: string): Promise<boolean> {
         const referenceSegments = reference.split(".");
         const firstSegment = referenceSegments[0];
         const fullSegments = ImportPathConverter.splitPath(fullPath);
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -649,14 +649,21 @@ export class ImportPathConverter {
      * descends the segments after it byte-exact from whichever match it takes.
      * So a nearer module of that name decides the whole reference, and one that
      * does not carry the rest of the chain breaks a reference whose full chain
-     * exists elsewhere - which is why matching the whole chain would confirm a
-     * spelling the compiler cannot read.
+     * exists elsewhere.
      *
      * Anything but one location refuses. Two or more mean a nearer namesake
      * shadows the target or clashes with it, and no relative spelling reaches
      * past it, so there is no conversion to offer rather than a longer one to
      * find. None means nothing here can place the module, which is exactly the
      * position this check exists to stop a conversion being written from.
+     *
+     * Two blind spots, both inherited from the search rather than added here.
+     * A namesake that exists only as an explicit `X := module` declaration is
+     * invisible while a folder answers first, because the declaration scan runs
+     * only when the folder phases found nothing. And a folder whose name
+     * differs from the reference only in case satisfies the search on a
+     * case-insensitive filesystem, so this confirms the casing it asked about
+     * rather than the casing on disk.
      */
     private async referenceResolvesTo(reference: string, fullPath: string, documentUri: vscode.Uri | undefined, projectVersePath: string): Promise<boolean> {
         const referenceSegments = reference.split(".");
@@ -686,7 +693,10 @@ export class ImportPathConverter {
      * The relative form of an absolute import, in the style the author wrote,
      * or null when there is none to give: a built-in module, an import that
      * was never absolute, a workspace with no `.uefnproject` to take the
-     * project path from, or a path that shortens to nothing.
+     * project path from, a path outside the project, a path that shortens to
+     * nothing, an importing file the Content root does not hold, or a
+     * shortened form that does not resolve back to the module the absolute
+     * path named.
      *
      * Shortened against the module the importing file sits in, which is the
      * dialect the compiler speaks: its own suggestions arrive spelled that way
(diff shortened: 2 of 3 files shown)


vukefn:
Follow-ups filed for what review turned up outside this seam: #428 (convert all reports declined imports as none found), #429 (nested plugin Content layout refuses every conversion), #430 (share first-segment lookups across a document-wide run). The multi-root blind spot stays with #396's H-14.

PR #431, opened 2026-08-15
fix(config): scope settings reads and derive the write target

### LINKED ISSUE #359: Settings facade: resource-scoped reads and inspect()-derived write target
## What

Settings access has no owner, and both halves of it are wrong as a class. No setting declares a `scope` and no read passes a resource, so folder-level values in a multi-root workspace are silently ignored; every configuration write hardcodes `ConfigurationTarget.Global`, so any workspace-level override permanently shadows every toggle in the status menu and the palette. Sixteen unscoped reads and twelve hardcoded-Global writes are not 28 mistakes; they are one missing module expressed 28 times across seven files.

## Where

- `package.json:158-327` — no `scope` key on any of the 22 properties.
- The 16 `getConfiguration("verseAutoImports")` reads, none passing a resource, across seven files: `extension.ts:76, 166, 194, 212, 261`; `CommandsHandler.ts:210`; `StatusBarHandler.ts:49, 102, 324, 389`; `DiagnosticsHandler.ts:155`; `ImportCodeLensProvider.ts:52, 57, 133`; `environment.ts:57`; `ImportCodeActionProvider.ts:27`.
- The 12 writes, all passing `vscode.ConfigurationTarget.Global`: `CommandsHandler.ts:213`; `StatusBarHandler.ts:148, 190, 199, 236, 250, 273, 346, 356, 366, 411, 421`.
- Three surfaces re-implement read-invert-write independently: `CommandsHandler.toggleConfig` (`:209-215`), the inline status-menu actions (`StatusBarHandler.ts:147-150` and five siblings), and the two submenus. The menu's Auto Import row (`StatusBarHandler.ts:144-151`) writes configuration directly instead of invoking the registered `verseAutoImports.toggleAutoImport` command, so the same toggle has two implementations maintained separately.

## Effect

A creator with two UEFN projects in one multi-root workspace cannot vary `behavior.importSyntax`, `general.autoImportScope` or any other behavior setting per project: a folder's `.vscode/settings.json` value is ignored without error, because undeclared scope defaults to `window`. Single-root layouts, the common case, are unaffected, which is what has kept this invisible.

On the write side, a user whose team repo commits `.vscode/settings.json` with `verseAutoImports.general.autoImport: false` clicks the status-menu Auto Import toggle: the write lands in user settings, the workspace value keeps winning, `updateDisplay()` re-reads the effective value, and the row still says Disabled. The toggle visibly does nothing, repeatably, with no error, in a status-menu-driven extension whose menu is the primary settings surface. Committed workspace settings in team UEFN projects are exactly where this lands.

Fixing the write half site-by-site means editing 12 places, and every future setting re-litigates scope and target; history shows it picks Global.

## Evidence

Manifest and all 16 call sites read and re-derived; all 12 write sites pass `vscode.ConfigurationTarget.Global` while every read is an effective-value `config.get`, so the read/write scope asymmetry is structural, not accidental. No comment anywhere documents Global as the intended target, and the Global writes arrived across ordinary feature commits (`642bfec`, `2ee87e3`, `15c63db`, `5b458b5`) with no stated reason. The only deliberate-scope marker in the lane (`extension.ts:23-24`) is about `[verse]` language blocks, not resource scope, and states the status quo rather than justifying it.

## Smallest fix

A settings facade of roughly 50 lines: typed keys, resource-scoped read, and an `inspect()`-derived write target (workspaceFolderValue -> WorkspaceFolder, workspaceValue -> Workspace, else Global), called from all 12 write sites. It turns both MEDIUM failures into two-function fixes and deletes the duplicated toggle logic. The per-row minimum, if taken separately: declare `"scope": "resource"` on the document-affecting settings and pass the document URI at the three read sites that hold one (`DiagnosticsHandler.ts:155`, `ImportCodeLensProvider.ts:133`, `ImportCodeActionProvider.ts:27`), or declare window scope explicitly and record why.

---
**Severity:** MEDIUM (A-14), MEDIUM (A-15).
**Source:** lane A platform review, rows A-14, A-15, verdict item 1. Full finding in `docs/review/lanes/A-platform/02-findings.md`, re-derived in `03-rulings.md`.

vukefn:
Closes #359

Settings access had no owner, and both halves were wrong as a class.

**Read half.** `package.json` contributed 22 settings and none declared a `scope`. An undeclared scope defaults to `window`, so a folder's `.vscode/settings.json` value was ignored without error in the multi-root workspace UEFN generates.

**Write half.** All 12 production write sites passed `vscode.ConfigurationTarget.Global`. With a workspace-level override committed in a team repo, clicking a toggle wrote to user settings, the workspace value kept winning, and the row still read Disabled — the control visibly did nothing.

The issue cites 16 unscoped reads; there were **24** in production by the time this was picked up.

## What changed

**`src/utils/settings.ts`** — one module owning scoping and write-target choice:

- `settingsFor(resource?)` returns `getConfiguration("verseAutoImports", resource)`. It owns *scoping*, not *reading* — call sites still write `.get<T>("key", default)` on what it returns. That is deliberate: `src/__tests__/configurationDefaults.test.ts:48` statically scans every non-test `.ts` for that literal shape and asserts each fallback matches the manifest default. Typed getters would have made that guard silently stop scanning while continuing to pass.
- `writeTargetFor()` walks `workspaceFolderValue -> workspaceValue -> globalValue`, the same ladder `extension.ts` already walked for reads, so the precedence exists once.
- `explicitSetting()` moved here from `extension.ts` unchanged.
- Scoping is by `Uri`, never `TextDocument`. A document carries a languageId, so passing one would silently switch on `[verse]` language-block resolution that has never applied here.

**Manifest** — an explicit scope on all 22 settings, 11 `resource` and 11 `window`, inserted without reformatting anything else.

`resource` is declared **only where a functional read passes a resource**. Declaring it where no read honours it would re-create the exact bug this fixes: advertising folder-configurability that is silently ignored.

**Reads** — the resource is threaded at the nine sites that hold one: `extension.ts` (hover), `DiagnosticsHandler` (the authoritative debounce read), `ImportCodeLensProvider.provideCodeLenses`, `ImportCodeActionProvider`, all three `ImportDocumentEditor` entry points, `ModuleVisibilityWriter`, and the two command surfaces, which scope to the active editor.

**Writes** — the five menu rows with a registered command now call `executeCommand`, matching how the same menu already invokes `optimizeImports`. That removes the second implementation of the Auto Import toggle and fixes a latent bug for free: those rows inverted a value snapshotted when the menu opened, where `toggleConfig` re-reads at click time. The three rows with no command write through the module.

## Two deviations from the issue, deliberate

**`general.autoImportScope` is `window`**, though #359 names it as a per-project example. It is read once per diagnostics event and evaluated against a snapshot of every open tab — it describes the editor's file set, not any one file — and the comment at `extension.ts` explicitly forbids moving that read per-uri. `resource` would be a promise the design cannot keep.

**`pathConversion.codeLensVisibility` and `experimental.useDigestFiles` are `window`**, because their functional reads hold no resource. Only the menu's display read could be scoped, and scoping the display while the behaviour stayed unscoped would show one value and use another.

## Review

Two rounds at `opus`, both driven against a real extension host rather than read off the page.

**Round 1 returned FAIL on a genuine Critical.** A **single-root** workspace registers one model as both its workspace and its folder configuration, so `inspect()` reports a `workspaceFolderValue` for window-scoped settings too. The first guard only checked that a resource had been passed, so it picked `WorkspaceFolder` for a window setting and `update()` threw — a regression against `main`, where the Global write merely did nothing. Toggling `experimental.useDigestFiles` in a single-root project that sets it would have raised `Action failed: ... does not support the folder resource scope`.

Fixed by deriving the folder target from the scopes the manifest declares (`RESOURCE_SCOPED`) rather than from `inspect()` alone, with a test asserting the set and the manifest agree — proven to fail in both directions, so they cannot drift into a thrown write.

Round 1 also exposed two coverage gaps by breaking the code and watching the suite stay green: unscoping the read in `addImportsToDocument`, and pointing a converted row at a dead command id. Round 2 exposed a third: the Critical's own fix was pinned only in the unit tests, with no case at the toggle a user clicks. All three now fail when broken.

**Round 2: PASS_WITH_SUGGESTIONS**, no Critical, no Important. It re-ran the probe that caught the Critical and confirmed the write now lands on `Workspace` and succeeds, and that the keys which legitimately reach `WorkspaceFolder` still do, across single-root and multi-root workspaces, with the overridden values rewritten in place at the level that held them.

Every guard in this PR was proven to detect its defect by removing the fix and watching only the matching test fail.

## Verification

```
$ npm run compile
> tsc -p ./
(clean)

$ npm test
Test Suites: 49 passed, 49 total
Tests:       1468 passed, 1468 total

$ npm run format:check

$ npx jest src/imports/__tests__
Test Suites: 10 passed, 10 total
Tests:       857 passed, 857 total
```

`src/imports/__tests__` was run in full for the repo's extractor/facade check: no pattern was added, moved or broadened in `ImportSuggestionExtractor`, nothing gained a path around `ImportHandler`, and no unrelated case changed shape.

CI is green, including `test-integration`, which exercises this change against a real multi-root extension host — the layer the unit suite's mocked `vscode` cannot reach.

`main` moved under this branch mid-flight (#423, #424, #427); it is merged in, and the one resolution that mattered kept `main`'s new lexer-based `ImportFormatter.commentStartIndex` from #424.

## Follow-ups, not fixed here

- **Quick-fix label can disagree with the text it inserts.** `behavior.importSyntax` is `resource`, but `ImportSuggestionExtractor` reads it unscoped, so in a multi-root workspace a folder overriding it gets a quick-fix *title* in the other syntax. The inserted text is correct — `ImportDocumentEditor` re-keys to the path and re-formats with its own scoped read — only the title is wrong, and it takes a folder-level override to appear. To be precise about what this is: it is not a pre-existing bug, it is a new capability with an incomplete edge, since before this change both reads were window-scoped and always agreed. Deferred because the fix touches `ImportHandler.extractImportSuggestions`'s signature. The cheap version is an **optional** trailing `resource?: vscode.Uri` forwarded to the two reads, which is source-compatible with every existing caller.
- **`ImportCodeLensProvider` leaks a hover timer in its own tests.** `npx jest src/imports/__tests__` intermittently prints "worker process has failed to exit gracefully"; the provider arms a `setTimeout` that only some of its tests dispose. Pre-existing and reproducible on this branch, which edits none of those tests and adds no timer.
- **`ModuleVisibilityWriter` pins `contentRoot` to `workspaceFolders[0]`** regardless of which folder the module lives in. Pre-existing, untouched here.
- **The extension-host suite's snooze coverage should be re-pointed.** `test-integration` runs in CI and **passes** on this branch, against the real multi-root `.code-workspace` harness. But it writes at `ConfigurationTarget.Workspace` while `commands.itest.ts` asserts on `.globalValue`, so now that writes are no longer unconditionally Global, its regression guard for "snooze must not write the setting" checks a level the code may no longer write to. It passes for a weaker reason than it used to.

### DIFF
--- a/changelog.d/359.fixed.md
+++ b/changelog.d/359.fixed.md
@@ -1,0 +1,1 @@
+**Settings**: toggles in the status menu and the command palette now take effect when a workspace or folder already sets the value, and per-folder settings apply in multi-root workspaces.
--- a/package.json
+++ b/package.json
@@ -182,24 +182,28 @@
       "title": "Verse Auto Imports",
       "properties": {
         "verseAutoImports.general.autoImport": {
+          "scope": "resource",
           "type": "boolean",
           "default": true,
           "description": "Automatically add missing imports when errors are detected",
           "order": 1
         },
         "verseAutoImports.general.diagnosticDelay": {
+          "scope": "window",
           "type": "number",
           "default": 1000,
           "description": "[DEPRECATED - Use autoImportDebounceDelay instead] Delay in milliseconds before processing diagnostics",
           "order": 2
         },
         "verseAutoImports.general.autoImportDebounceDelay": {
+          "scope": "window",
           "type": "number",
           "default": 3000,
           "description": "Debounce delay in milliseconds before auto-importing. Waits for this duration after the last diagnostic change before triggering auto-imports (while you're typing).",
           "order": 3
         },
         "verseAutoImports.general.autoImportScope": {
+          "scope": "window",
           "type": "string",
           "enum": [
             "allFiles",
@@ -211,6 +215,7 @@
           "order": 4
         },
         "verseAutoImports.behavior.importSyntax": {
+          "scope": "resource",
           "type": "string",
           "enum": [
             "curly",
@@ -221,18 +226,21 @@
           "order": 10
         },
         "verseAutoImports.behavior.preserveImportLocations": {
+          "scope": "resource",
           "type": "boolean",
           "default": true,
           "description": "When true, keeps existing imports where they are and only adds new imports at the top. When false, consolidates all imports at the top of the file.",
           "order": 11
         },
         "verseAutoImports.behavior.sortImportsAlphabetically": {
+          "scope": "resource",
           "type": "boolean",
           "default": true,
           "description": "When true, sorts imports: absolute paths and dotted references alphabetically, while bare local imports keep their relative order and precede dotted ones to preserve Verse's top-down using resolution. When false, preserves the original order.",
           "order": 12
         },
         "verseAutoImports.behavior.importGrouping": {
+          "scope": "resource",
           "type": "string",
           "enum": [
             "none",
@@ -244,6 +252,7 @@
           "order": 13
         },
         "verseAutoImports.behavior.emptyLinesAfterImports": {
+          "scope": "resource",
           "type": "integer",
           "default": 1,
           "minimum": -1,
@@ -252,12 +261,14 @@
           "order": 14
         },
         "verseAutoImports.behavior.ambiguousImports": {
+          "scope": "window",
           "type": "object",
           "default": {},
           "description": "Mappings for class names that appear in multiple modules. Format: {\"className\": \"preferredModulePath\"}",
           "order": 15
         },
         "verseAutoImports.behavior.multiOptionStrategy": {
+          "scope": "resource",
           "type": "string",
           "enum": [
             "quickfix",
@@ -270,6 +281,7 @@
           "order": 16
         },
         "verseAutoImports.behavior.digestImportPrefixes": {
+          "scope": "window",
           "type": "array",
           "items": {
             "type": "string"
@@ -283,24 +295,28 @@
           "order": 17
         },
         "verseAutoImports.quickFix.sortAlphabetically": {
+          "scope": "resource",
           "type": "boolean",
           "default": false,
           "description": "When true, sorts quick fix options alphabetically. When false, preserves original order.",
           "order": 20
         },
         "verseAutoImports.quickFix.showDescriptions": {
+          "scope": "resource",
           "type": "boolean",
           "default": false,
           "description": "Show descriptive text in quick fix menu items (e.g., 'class from /Fortnite.com/Devices')",
           "order": 21
         },
         "verseAutoImports.pathConversion.enableCodeLens": {
+          "scope": "resource",
           "type": "boolean",
           "default": true,
           "description": "Show CodeLens actions above import statements to convert between absolute and relative paths",
           "order": 30
         },
         "verseAutoImports.pathConversion.codeLensVisibility": {
+          "scope": "window",
           "type": "string",
           "enum": [
             "hover",
@@ -311,6 +327,7 @@
           "order": 31
         },
         "verseAutoImports.pathConversion.codeLensHideDelay": {
+          "scope": "window",
           "type": "number",
           "default": 1000,
           "minimum": 500,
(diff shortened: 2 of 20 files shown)


vukefn:
Follow-ups filed from this change: #433 (quick-fix label vs inserted text, the incomplete edge this PR introduces), #434 (snooze guard asserts on globalValue), #435 (CodeLens hover timer leak, pre-existing), #436 (visibility writer pins workspaceFolders[0], pre-existing).

PR #441, opened 2026-08-15
fix(imports): resolve phase-2 module locations with the shared matcher

### LINKED ISSUE #395: Phase 2: delete the bespoke location matching and surface the 100-file cap

vukefn:
Closes #395

Phase 2 of the module search decided where a module lived by testing whether the declaring file's directory string ended with the rest of the reference. That model cannot see nesting, and the correct one — segment-aligned, chain-aware, unit-tested — already existed in `resolveModuleLocations`, which is what the cache path uses.

## What changed

**One matcher, imported twice.** `searchExplicitModuleDefinitions` now reads each declaration with its enclosing module chain through `findExplicitModuleDeclarations` and matches through `resolveModuleLocations`. The directory-suffix rule, the Content-prefix arithmetic and the `startsWith`/`endsWith` string work are gone; `toContentRelativeDir`, exported for this, does that job inside the shared matcher.

A module declared inside another is an ordinary path scope (`SemanticAnalyzer.cpp:4416-4420` allows a module definition at module scope; resolution walks the segments through those scopes at `:6238-6260`). So `Outer.Inner` names the inner module and resolves to the declaring file's directory — where before it was reported missing, because directory `X` does not end with `Outer`. The bare name `Inner`, reachable only through `Outer`, now resolves to nothing instead of emitting the declaring file's directory and building `/proj/X/Inner`, a path nothing declares, written with `isAmbiguous: false`.

**`MODULE_DECLARATION`'s specifier body is narrowed to `[^<>]+`.** This is required by the reuse rather than adjacent: the deleted `buildModuleDefinitionRegex` carried this narrowing and the shared reader did not. With `[^>]+` a `<` that opens nothing — a comparison operator — runs to the `>` of a later declaration's specifier, so `if (Inventory < MaxSlots): ... Helpers<public> := module:` reported a module named `Inventory` and missed `Helpers` entirely. No specifier nests, so nothing legitimate is lost; the body still spans lines, because a `scoped{...}` list may be wrapped. The visibility writer shares this pattern and gains the same correction — for it a misread name means editing a declaration that does not exist.

**The scan's file cap is named and reported.** `DECLARATION_SCAN_FILE_LIMIT` stays at 100. The glob asks for one file past it and never reads that one, so a project holding exactly the limit is told apart from one holding more — asking for the limit itself makes those two answers identical, which is the conflation the signal exists to end. `findModuleLocations` returns `{ locations, truncated }`.

Both entry points fail closed on `truncated`, following the direction PR #427 set: `convertToFullPath` declines with a message naming the cap rather than the existing "Could not find module", and `referenceResolvesTo` refuses ahead of its location count, which makes explicit the limit its doc comment already recorded.

**Distrust is scoped to the phase that hit the cap.** `truncated` covers the declaration scan's own answer and an empty result — where the module may be declared in the part left unread — but not an answer the folder search reached on its own. That phase reads the project's directories for itself under a cap fifty times higher, and a folder module has no declaration for the cache to hold either, so vetoing it would have switched conversion off for the commonest case that reaches this code at all.

The cap is reported once per run rather than once per import, since the condition belongs to the project: converting a whole document raised one notification per line.

## Behaviour this removes

A project holding more than 100 `.verse` files no longer gets a conversion for a module only the declaration scan could place. It got one before, resolved from whichever files the glob happened to return — the silent wrong module the ticket describes. Everything the folder phases place is unaffected at any project size.

## What this knowingly leaves open

The cap cannot be made both safe and useful by a rule, and the review established both horns of that. Distrusting the folder search after a partial declaration scan refuses every folder module in a project past 100 files. Trusting it leaves one narrow route to the ticket's own harm: a module declared in the unread remainder, where a folder namesake answers instead and is written unambiguously. The same project one file smaller answers differently.

The rule takes the second, because a folder module has no declaration anywhere for the scan or the cache to hold, so the first turns the feature off for the commonest thing that reaches this search at all. The comment at the assignment states this as a trade rather than asserting the gap away. **Raising the cap is what shrinks the exposure; no rule at 100 can.** That is the ticket's own deferred "the number itself should scale or page", and it now has evidence behind it.

## Not fixed here

- **A quoted identifier whose suffix contains a dot** (`Inv'a.b'`) no longer resolves. `ProjectPathNode.fullPath` encodes the chain as a dotted string and `resolveModuleLocations` recovers it with `split(".")`. This is a regression this branch ships, not only a shared gap: the deleted `buildModuleDefinitionRegex` escaped the name and matched it. The cache path builds `fullPath` the same way (`ProjectPathScanner.ts:414`, `:457`), so the fix belongs in the node contract for both producers. Filed separately.
- **After a cap refusal the command layer still adds a generic line** — "No relative imports found to convert." for a document that holds three. The converter reports the cap once per run, but `CommandsHandler` cannot tell "refused for the cap" from "nothing to convert" without a decline reason on the result. The reverse direction gets only the generic line. The pattern predates this branch; this refusal path makes it common.
- **`FOLDER_SCAN_FILE_LIMIT` (5000) has no truncation signal at all** — the same shape this ticket closed for phase 2, one phase over.
- The ticket's "the number itself should scale or page" — deliberately out of this diff; the number is unchanged.

## Review

Two rounds, fresh context each, at `models.review`.

**Round 1 — FAIL** on a real Critical I had reasoned my way into: I scoped the truncation veto to the whole search, and the reviewer measured what that costs. Folder modules never hit the declaration cache, so conversion switched off for them in any project past 100 files, including answers phase 3 reached completely on its own. Fixed in `b17715a`, which also dropped advice to enable a setting that defaults to on, collapsed one notification per import into one per run, made the cap boundary exact, and marked a throwing scan as a partial read.

**Round 2 — FAIL** with no Critical, on the comment I wrote justifying the narrowed rule: it claimed the residual gap could not happen, and the reviewer built the project that reproduces it. Fixed in `5c5da32` — the rule stands, the comment now states the trade and names raising the cap as the only thing that shrinks it. Its remaining `Important` on the command layer's generic line is listed above and filed rather than fixed, since it needs a decline reason threaded to `CommandsHandler`.

## Verification

```
$ npm run compile
> tsc -p ./

$ npm test
Test Suites: 50 passed, 50 total
Tests:       1513 passed, 1513 total

$ npm run format:check
Checking formatting...
```

Every new test was proven to detect its defect by taking the fix away and watching it fail: `fullPath: chain.join(".")` reduced to `declaration.name` (the nested reference misses, the bare inner name fabricates a location, the entry point returns null), the specifier body widened back to `[^>]+` (the comparison operand is read as the declared module), the truncation signal forced to `false` (a 101-file project writes a sample's answer with `isAmbiguous: false`), and the scoping rule reduced to `declarationScanReadPart` (the folder search's own answer is vetoed).

### DIFF
--- a/changelog.d/395.fixed.md
+++ b/changelog.d/395.fixed.md
@@ -1,0 +1,1 @@
+**Import path conversion**: converting a `using` to its absolute path now resolves modules declared inside other modules, and declines rather than guessing when the project is too large for the declaration scan to read in full.
--- a/changelog.d/395.fixed.md
+++ b/changelog.d/395.fixed.md
@@ -1,1 +1,1 @@
-**Import path conversion**: converting a `using` to its absolute path now resolves modules declared inside other modules, and declines rather than guessing when the project is too large for the declaration scan to read in full.
+**Import path conversion**: modules declared inside other modules now convert in both directions, and a project too large for the module search to read in full is told so once rather than offered a path the search cannot stand behind.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -1,10 +1,11 @@
 import * as vscode from "vscode";
 import * as path from "path";
 import { logger } from "../utils";
-import { maskCommentsAndStrings } from "../utils/verseText";
 import { ProjectPathHandler } from "../project";
 import { ProjectPathCache } from "../services";
-import { resolveFolderModuleLocations, toContentRelativeDir } from "../services/moduleLocationLookup";
+import { buildProjectIndexes, resolveFolderModuleLocations, resolveModuleLocations, toContentRelativeDir } from "../services/moduleLocationLookup";
+import { ProjectPathNode } from "../types";
+import { findExplicitModuleDeclarations } from "../visibility/moduleDeclarations";
 import { ImportFormatter } from "./ImportFormatter";
 import { LINE_SPLIT, scanConvertibleImports } from "./ImportScanner";
 
@@ -32,13 +33,37 @@ interface ImportConversionResult {
     line?: number;
 }
 
+/**
+ * Where a module could be, and whether the search that answered was complete.
+ *
+ * `truncated` says the declaration scan stopped at its file cap, so `locations`
+ * holds whatever a sample of the project attested to. A location absent from it
+ * may still exist, and a single entry is no evidence that the name is
+ * unambiguous, so nothing may be written from a truncated answer.
+ */
+interface ModuleLocationSearch {
+    locations: string[];
+    truncated: boolean;
+}
+
 const CONTENT_FOLDER = "Content";
 
 /**
+ * How many `.verse` files the explicit-declaration scan reads. Two orders of
+ * magnitude below FOLDER_SCAN_FILE_LIMIT because this phase opens and lexes
+ * every file it scans, where the folder search reads none.
+ *
+ * It stays small because a project past it answers from the declaration cache
+ * instead: this scan runs only once the cache has missed, and above the limit it
+ * reports itself truncated rather than answering from a sample.
+ */
+const DECLARATION_SCAN_FILE_LIMIT = 100;
+
+/**
  * How many `.verse` files the project-wide folder search enumerates. Far above
- * the explicit-declaration scan's 100 because this one reads no file contents:
- * it derives folder modules from the paths alone, so a file costs a string
- * split rather than a read.
+ * DECLARATION_SCAN_FILE_LIMIT because this one reads no file contents: it
+ * derives folder modules from the paths alone, so a file costs a string split
+ * rather than a read.
  */
 const FOLDER_SCAN_FILE_LIMIT = 5000;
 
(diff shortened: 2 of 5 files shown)


vukefn:
Follow-ups filed from this lane, all deliberately out of #395's scope:

- #442 — a quoted identifier holding a dot no longer resolves. A regression this branch ships, not only a shared gap: the deleted `buildModuleDefinitionRegex` escaped and matched the whole name. The fix belongs in the `ProjectPathNode` contract, which both producers share.
- #443 — after a cap refusal the command layer still says the document had nothing to convert.
- #444 — the folder search has no truncation signal of its own, which is the guarantee this PR's rule leans on past 5000 files.
- #445 — the cap should scale or page, with the review's evidence for why a larger cap is the only thing that shrinks the gap documented here.

PR #447, opened 2026-08-15
fix(imports): place a file under a Content root at any depth

### LINKED ISSUE #429: Nested plugin Content layout refuses every relative-path conversion
## What

`enclosingModulePath` only accepts a `Content` folder at the top of the workspace folder, so a file in the nested-plugin layout cannot be placed and every relative-path conversion in it is refused.

## Where

- `ImportPathConverter.ts:579-597` — `enclosingModulePath`; the `startsWith("Content/")` test at `:594` is against a path relative to the workspace folder root.
- `ImportPathConverter.ts:765-769` — the placeability guard added by #392, which turns "cannot place the file" into a refused conversion.
- `ProjectPathHandler.ts:41-46` — documents supporting the `<project>/Plugins/<Name>/Content` layout.

## Effect

With the workspace opened at the project root and a file at `Plugins/MyGame/Content/Systems/Feature.verse`, "Use relative path" reports `"Import cannot be converted to relative path."` and does nothing. Before #392 the same conversion produced `using. Gadgets`, which was the right answer for that layout.

The refusal is correct under #392's fail-closed rule — the search genuinely cannot place the file, so nothing can verify the reference it would emit — but the cause is the `Content` boundary being read at one fixed depth, not a real absence of scope. A layout the codebase claims to handle silently lost a feature.

## Relation to existing work

This is the same `Content`-boundary rule as H-08 in #396 ("the `Content` boundary", `ImportPathConverter.ts:589, 594` among the cited sites). Filed separately because the user-visible effect is new — #396 describes a silent scope fallback, and after #392 the same root cause presents as a refused conversion instead. Fixing H-08 to find the nearest enclosing `Content` root rather than one at a fixed depth would close both; if that is how #396 is taken, close this as a duplicate.

---
**Source:** review of PR #427 (#392), round 2 finding 2, confirmed by probe.

vukefn:
Closes #429

`enclosingModulePath` and `searchImplicitModules` both read the `Content` boundary against a path relative to the workspace folder root, so a `Content` directory anywhere below depth zero placed no file at all. With the workspace opened at the project root and a file at `Plugins/MyGame/Content/Systems/Feature.verse`, phase 1 of the module search returned before walking, the scope came back null, and the placeability guard from #392 turned both into the refusal the issue reports.

One private helper now locates the root along the whole path from the workspace folder down to the file, and the two sites read it. The **shallowest** `Content` wins rather than the nearest, so a folder module named `Content` below the root is not mistaken for the root and its files are not addressed one scope too low.

## The layout is the current one, not a legacy one

`ProjectPathHandler`'s doc comment calls the nested layout "UEFN's older project layout". A live check against 40 real `.uefnproject` projects says otherwise: **25 are nested `Plugins/<Name>/Content`, 15 are flat `Content`, and none carry both**. A current project file (`fileVersion 15`, `compatibilityVersion 35.10`) puts `.uefnproject` at the project root with Verse under `Plugins/<Name>/Content` and a single root plugin. So this was not a fringe layout: it is the majority shape, and "Use relative path" was refused in all of it whenever the workspace was opened at the project root.

Every one of the 40 projects declares exactly one plugin, always `bIsRoot`. The multi-plugin case is bounded anyway by the existing outside-the-project guard in `convertFromFullPath`, which refuses any import whose prefix does not match `projectVersePath`, so a second plugin's imports cannot be silently re-addressed under this project's path.

## Only the root plugin's Content root is admitted

Placing a file under whatever `Content` sits on its path would have gone too far, and the first review round caught it: only the root plugin's Content answers to `bindings.projectVersePath`, so a file under `Plugins/<Other>/Content` would have been scoped and addressed under the wrong package — a refusal traded for a silent wrong write, which is the failure class #392 exists to prevent. `referenceResolvesTo` cannot catch that, because it re-derives the scope from the same project path and so confirms the reference it was handed.

A Content root **below** the workspace folder's top level is therefore admitted only when its parent is the root plugin's declared name and its grandparent is `Plugins`, read from the project file's `plugins` array — which was parsed already and had no readers. A Content **directly under** the workspace folder is taken as before, without that evidence: the workspace is then opened at the thing holding it, and requiring evidence there would drop the flat layout that already worked. In the sample, all 15 flat projects have no plugin directory at all, and 24 of the 25 nested ones have exactly one plugin Content directory whose name equals the declared root plugin's.

## Known residual, owned by no ticket yet

`workspaceScanTargets` still globs `Content/**/*.verse` from the workspace folder, so module-search phases 2 and 3 stay blind to a nested project; `ProjectPathCache.workspaceIsContent()` and `toContentRelativeDir` read the same fixed depth, and `ModuleVisibilityWriter.findContentRoot` reads it for a different feature. `findModuleLocations` runs phase 1 first and falls through only when it found nothing, so fixing phase 1 restores the reported behaviour — but "Use full path" on a module only the project-wide scan could reach still fails in this layout, fail-closed rather than wrong.

This is **not** covered by #396: its H-08 is the byte-exact *case* rule and H-14 is `workspaceFolders[0]`, and neither is the *depth* rule. A follow-up issue should own it rather than this PR widening into it.

## Review

Two rounds at opus, both probed rather than read. Round 1 returned FAIL on one Critical, which is what produced the root-plugin gate above: the first attempt placed a file under whatever `Content` sat on its path, which would have addressed a non-root plugin's files under the wrong package — a refusal traded for a silent wrong write. Round 2 verified that closed against `origin/main`, confirmed both directions and the round trip, and returned PASS_WITH_SUGGESTIONS.

The remaining suggestions were taken except one: moving the plugin check onto the path segments the caller already holds, instead of building an absolute path and taking it apart. Declined because it would narrow real behaviour — reading the absolute path is what lets a workspace opened at `<project>/Plugins` or at `<project>/Plugins/<Name>` still place its files, where segments relative to the workspace folder cannot see the `Plugins` level at all.

## Tests

Thirteen cases in a new `under a nested plugin Content root` block, pinned at `convertFromFullPath` — the entry point `CommandsHandler.convertToRelativePath` calls and the one that returned null. They cover the bare-name and multi-segment forms, a file at the nested root itself, the round trip back through `convertToFullPath`, the shallowest-Content rule, a file no Content root holds, a non-root plugin's Content refused in both directions, a project declaring no root plugin, a Content nested under something other than `Plugins`, a declared plugin name differing from the directory only in case, an unrelated `Content` beside `Plugins`, and workspace-at-`Content` winning over a `Content` nested below it.

Every assertion was proven to detect its defect by taking the code away rather than by passing: five of the six original cases fail against the unfixed source (the sixth is a negative case, which correctly refuses either way); all three gate cases fail with the root-plugin check disabled; and the two halves of the gate's own rule were mutation-tested separately — dropping the `Plugins` grandparent check fails one case, replacing the case fold with the identity fails three.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Test Suites: 50 passed, 50 total
Tests:       1533 passed, 1533 total
Snapshots:   0 total
Time:        7.967 s, estimated 11 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/429.fixed.md
+++ b/changelog.d/429.fixed.md
@@ -1,1 +1,1 @@
-**Path conversion**: "Use relative path" now works in the UEFN project layout that nests Verse under `Plugins/<Name>/Content`, where every conversion in such a file was refused.
+**Path conversion**: imports convert in both directions again for a project that nests its Verse under `Plugins/<Name>/Content`, the UEFN layout where every conversion was refused when the workspace was opened at the project root.
--- a/changelog.d/429.fixed.md
+++ b/changelog.d/429.fixed.md
@@ -1,0 +1,1 @@
+**Path conversion**: "Use relative path" now works in the UEFN project layout that nests Verse under `Plugins/<Name>/Content`, where every conversion in such a file was refused.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -249,6 +249,46 @@ export class ImportPathConverter {
     }
 
     /**
+     * Where the project's Content root sits inside the workspace folder and
+     * where the file sits under that root, both as forward-slash paths with
+     * `""` for the enclosing directory itself, or null when no directory on
+     * the way down to the file is a Content root.
+     *
+     * The root is looked for along that whole path rather than read at a fixed
+     * depth, because UEFN's shipped layout puts Verse under
+     * `<project>/Plugins/<Name>/Content`: a workspace opened at the project
+     * root reaches Content three levels down, and a boundary that only admits
+     * depth zero places no file in such a project at all.
+     *
+     * The shallowest Content wins rather than the nearest. A Content below the
+     * root is an ordinary folder module, and taking that one as the root would
+     * address every file under it one scope too low.
+     */
+    private static placeUnderContentRoot(workspaceFolderPath: string, documentPath: string): { contentRootRelative: string; fileDirRelative: string } | null {
+        const relativeFilePath = path.relative(workspaceFolderPath, documentPath).replace(/\\/g, "/");
+
+        // path.relative climbs out with `..` for a file the folder does not
+        // hold, and a Content segment past that point belongs to some other
+        // tree - the placement has to be inside the folder to mean anything.
+        if (relativeFilePath.startsWith("../")) return null;
+
+        const fileDir = path.dirname(relativeFilePath).replace(/\\/g, "/");
+        const dirSegments = fileDir === "" || fileDir === "." ? [] : fileDir.split("/");
+
+        if (path.basename(workspaceFolderPath) === CONTENT_FOLDER) {
+            return { contentRootRelative: "", fileDirRelative: dirSegments.join("/") };
+        }
+
+        const rootIndex = dirSegments.indexOf(CONTENT_FOLDER);
+        if (rootIndex < 0) return null;
+
+        return {
+            contentRootRelative: dirSegments.slice(0, rootIndex + 1).join("/"),
+            fileDirRelative: dirSegments.slice(rootIndex + 1).join("/"),
+        };
+    }
+
+    /**
      * Appends the locations where a folder of this name sits, searched outwards
      * from the current file: siblings first, then each ancestor directory, then
      * the Content root. A folder is an implicit module, so its presence is the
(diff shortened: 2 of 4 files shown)


vukefn:
The residual this PR leaves is now tracked as #448: the same fixed-depth `Content` rule in `workspaceScanTargets`, `ProjectPathCache.workspaceIsContent`, `toContentRelativeDir` and `ModuleVisibilityWriter.findContentRoot`. Not covered by #396, which is the case rule and the multi-root rule.

Issue #436, opened 2026-08-15
Module visibility writer always resolves the content root from workspaceFolders[0]


vukefn:
## What

`ModuleVisibilityWriter` resolves the content root from `vscode.workspace.workspaceFolders?.[0]` regardless of which folder the module being made public actually lives in. In a multi-root workspace it can read and write the wrong folder's definitions file.

## Where

- `src/visibility/ModuleVisibilityWriter.ts:122` — `workspaceFolder` is pinned to `workspaceFolders?.[0]`
- `src/visibility/ModuleVisibilityWriter.ts:143` — `findContentRoot(workspaceFolder)`
- `src/visibility/ModuleVisibilityWriter.ts:330-331` — `definitionsUri(contentRoot)`, which now scopes the `moduleVisibility.definitionsFileName` read to that same pinned root
- `src/visibility/ModuleVisibilityMessage.ts:35,38` — `request.targetPath` / `request.importerPath` are plain strings, so the real location is available but never used to pick the folder

## Effect

A creator with the multi-root workspace UEFN generates invokes the make-module-public quick fix on a module under the second or third folder. The writer looks for `_definitions.verse` under folder 0's content root. It either edits the wrong project's definitions file or reports that it cannot find one, in both cases silently with respect to the folder the user was actually working in.

Single-root layouts are unaffected, which is what has kept this invisible.

Two things follow from the same pin: the wrong file is chosen, and `moduleVisibility.definitionsFileName` — declared `"scope": "resource"` — resolves against folder 0 rather than the module's own folder, so a per-folder override of the filename is ignored too.

## Why it is filed separately

Pre-existing. #359 scoped the settings read to `contentRoot`, which is correct as far as it goes, but deliberately did not touch where `contentRoot` comes from — that is this issue.

## Smallest fix

Derive the workspace folder from the module the request names, via `vscode.workspace.getWorkspaceFolder(uri)` on the resolved target path, and fall back to `workspaceFolders[0]` only when that answers nothing.

Note for whoever takes it: the `vscode` mock's `getWorkspaceFolder` matches on path prefix, returns the *first* containing folder where real VS Code returns the innermost, and answers `undefined` for a URI equal to a folder root. Its own doc comment says a multi-root or nested-folder test needs more than that, so the mock will need work before it can express this case.

## Acceptance

A test with two workspace folders, where a module under the second one is made public, writes to the second folder's definitions file and honours that folder's `moduleVisibility.definitionsFileName`.

Issue #376, opened 2026-08-13
Project identity: one owner emitting one event, replacing three watchers


vukefn:
## What

Project identity is held in three components with no ordering contract between them, and each registers its own `**/*.uefnproject` string-glob watcher. In the UEFN-generated multi-root workspace the project file sits outside every workspace folder, where a string-glob watcher cannot deliver events, so none of the three can fire where users actually are. Discovery itself only works in that shape through undocumented `findFiles` behaviour, with no test that would notice it breaking.

## Where

- ProjectPathHandler.ts:192; ProjectPathCache.ts:255; AssetsDigestParser.ts:298 — the three string-glob registrations.
- ProjectPathHandler.ts:54-56 — identity is cached on success only.
- ProjectPathCache.ts:251-254 ("the file appearing is the signal that a rebuild can finally succeed") and AssetsDigestParser.ts:295-297 — comments promising watcher-driven recovery.
- AssetsDigestParser.ts:44-50 — states in writing that nothing contracts the ordering between its refresh and the handler's cache clear.
- ProjectPathHandler.ts:107-108 — the parent walk, passing an out-of-workspace `RelativePattern` base.
- test-fixtures/integration-workspace/VerseAutoImportsIT.code-workspace — folder 0 is `Project/Content`; the `.uefnproject` sits in `Project/`, outside every folder.
- commands.itest.ts:43-47 and activation.itest.ts:32-41 — the only extension-host tests nearby; they assert that the rebuild command resolves without throwing and that activation survived.

## Effect

A `.uefnproject` change mid-session — a rename, a Verse-path change, UEFN rewriting bindings — is never observed. `ProjectPathHandler` serves the stale identity for the rest of the session, the cache never rebuilds, the assets parser keeps the old project's names, and path-conversion features write the old `projectVersePath` into new `using` lines. Create-after-startup partially recovers by accident, because a failed discovery is not cached and every later `getProjectName()` redoes the walk. The comments promising watcher-driven recovery describe a mechanism that cannot run in this workspace shape; they hold only in single-root workspaces where the project file is inside the folder.

Discovery is the second half of the same exposure: the parent walk works today, but on behaviour the documentation contradicts. A platform change to the documented semantics would silently kill project identity, the scan, the assets digest and the cache in the flagship workspace, with every unit test still green.

## Evidence

[...]

PR #460, opened 2026-08-15
fix(project): resolve the project file from the folder its file sits in

### LINKED ISSUE #450: Project path handler answers for the first .uefnproject in the workspace, not the file's own
## What

`ProjectPathHandler.findAndParseProjectFile` iterates every workspace folder but returns on the **first** `.uefnproject` it finds, caching that project file for the life of the handler. Every consumer of `getProjectVersePath()`, `getProjectName()` and `getProjectModules()` therefore answers for whichever project happens to sit in the lowest-indexed folder, regardless of which project the file being worked on belongs to.

## Where

- `src/project/ProjectPathHandler.ts:64-93` — the `for (const folder of workspaceFolders)` loop returns inside the first iteration that matches, and assigns `this.cachedProjectFile` / `this.projectVersePath` / `this.projectName` as it goes
- `src/project/ProjectPathHandler.ts:95-96` — the parent-directory fallback then walks up from `workspaceFolders[0]` specifically
- `src/project/ProjectPathHandler.ts:144-151` — `getProjectVersePath()` short-circuits on the cached value, so the first answer is the only answer

## Effect

In a multi-root workspace holding two UEFN projects, a file in the second project resolves its module paths against the first project's `projectVersePath`. In `ModuleVisibilityWriter.buildEdit`, `toContentSegments(request.targetPath, projectVersePath)` then refuses the request with `'<path>' is outside the project '<other project>'`, which is a confusing message: the module is inside its own project, just not the one that won the race.

Single-project workspaces are unaffected whatever their folder count, which is what has kept this invisible.

## Relationship to #436

Found while fixing #436 and deliberately left out of that diff. #436 makes the *content root*, the settings scope and the declaration-scan subtree follow the module being edited. This issue is the remaining half: the *project path* those segments are computed against still comes from folder 0. Both must be folder-scoped before a two-project workspace works end to end.

## Smallest fix

Key the resolution on the resource being worked on rather than caching one project file per handler: resolve upward from the file's own directory, or from `getWorkspaceFolder(uri)`, and cache per folder rather than per handler. `clearCache()` and the `.uefnproject` watcher would need to clear the whole map.

Worth deciding first whether the cache should be per workspace folder or per resolved project root — the parent-directory walk at lines 95-133 means two folders can legitimately share one project file.

## Acceptance

With two workspace folders each holding a different `.uefnproject`, `getProjectVersePath()` asked about a file in the second folder answers the second project's Verse path.

vukefn:
Closes #450

## What was wrong

`ProjectPathHandler` searched every workspace folder for a `.uefnproject` but returned on the first hit and cached it on the instance, so `getProjectVersePath`, `getProjectName`, `getRootPluginName` and `getProjectModules` answered for whichever project sat in the lowest-indexed folder for the life of the handler. In a multi-root workspace holding two UEFN projects, a file in the second project resolved against the first project's Verse path, and `ModuleVisibilityWriter` then refused with `'<path>' is outside the project '<other project>'` — a message naming the wrong project.

This is the remaining half of #436 (PR #452), which made the content root, the settings scope and the declaration-scan subtree follow the requesting document but left the project path those segments are measured against coming from folder 0.

## What changed

**`src/project/ProjectPathHandler.ts`** — every accessor takes an optional resource and answers for the project that owns it, caching per workspace folder instead of per handler. The cache is keyed on the folder searched from rather than the project file found, because the key has to be known before the search that would reveal the file; two folders under one project therefore hold equal entries, which costs a second parse and never a wrong answer. `clearCache()` empties the whole map. A call naming no resource keeps the whole-workspace search, which is what the activation-time readers need.

**`src/imports/ImportPathConverter.ts`** — the two `getProjectVersePath` reads and the two `getRootPluginName` reads pass the document being converted, so the path and the plugin name always describe one project where they meet in `placeUnderContentRoot`.

**`src/services/ProjectPathScanner.ts`** — `scanProject` passes the folder it was handed.

**`src/visibility/ModuleVisibilityWriter.ts`** — the content-root block moved above the project-path read, and the path is read for the folder that won rather than for `sourceUri`. `locateContentRoot(requestedFolder) ?? locateContentRoot(firstFolder)` can settle on a folder other than the requesting document's, and scoping the path to `sourceUri` alone would then name a project the edit is not being written into.

**`src/project/__tests__/ProjectPathHandler.test.ts`** — new; the class had no suite.

## Behaviour changes beyond the fix

Three, all deliberate, none pinned by an existing test:

1. A `.uefnproject` in the file's own folder that will not parse now answers nothing, where the old code fell through to the first folder's project. Answering some other project's paths is the bug this ticket is about, so refusing is the correct answer here. Pinned by a test.
2. Where a resource's own folder holds no project but an ancestor directory does, that ancestor now outranks an unrelated sibling workspace folder that has one. The old code preferred the sibling. The file is genuinely inside its ancestor's project, so this is the better answer — but it is a change rather than a preservation, and the plan claimed preservation until review caught it.
3. Refusal ordering in `ModuleVisibilityWriter`: when no `Content` folder is present, "no 'Content' folder was found" now wins over "no .uefnproject file was found", over "is outside the project" and over "already reachable". No refusal was lost and no success became a refusal — only which message wins when several conditions hold at once.

## Verification

`npm run compile`

```
> tsc -p ./
```

`npm test`

```
> jest --verbose

Test Suites: 53 passed, 53 total
Tests:       1593 passed, 1593 total
Snapshots:   0 total
Time:        11.972 s
Ran all test suites.
```

`npm run format:check`

```
> prettier --check "**/*.ts"

Checking formatting...
```

The regression test was proved to detect the defect: with the resolution neutered but the signature intact, 6 of the 8 new tests fail with exactly the reported symptom (asked about the second project, answering `/mygame@fortnite.com/firstgame`). The 2 that still pass are the workspace-wide cases, which correctly do not depend on scoping.

`npx jest src/imports/__tests__` was also run in full and read whole, per the repo's extractor-order check: 10 suites, 908 tests, nothing changed shape.

## Review

One round at opus: `PASS_WITH_SUGGESTIONS`, 0 Critical, 3 Important, 4 Suggestions. Fixed in the second commit:

- `getRootPluginName` had gained the resource parameter but kept two callers passing nothing, leaving a folder-scoped Verse path meeting a workspace-scoped plugin name.
- The changelog fragment claimed the project *name* was folder-scoped; its only scoped caller is reachable only through a code path this change leaves on folder 0. Trimmed to what the change actually reaches.
- The claim that the new search "preserves every answer the old code gave" was falsifiable; it is now recorded as behaviour change 2 above.

Also taken: narrowed `readProjectFileIn` to the `WorkspaceFolder` it is actually called with, and stopped the all-folders loop re-searching the folder already searched as the scope.

Left as reported, not fixed: the extension has no `onDidChangeWorkspaceFolders` invalidation (pre-existing, and the per-folder map is no more exposed to it than the single cached file was), and a workspace with no project file still costs a repeated search because misses are not cached.

## Not in this change

- `AssetsDigestParser.getAssetsDigestPath` and `ProjectPathCache.loadFromStorage` are activation-time and workspace-global, with no resource anywhere in their call chains, as is `ProjectPathCache.rebuildCache`'s hardcoded `workspaceFolders[0]`. Giving those a project identity is #376's job.
- `ImportPathConverter` constructs its own second `ProjectPathHandler` that never gets a file watcher, so its cache outlives a project-file edit. Pre-existing and separate; worth its own ticket, and this change raises the stakes since that instance is now the main consumer of the folder-scoped path.

### DIFF
--- a/changelog.d/450.fixed.md
+++ b/changelog.d/450.fixed.md
@@ -1,1 +1,1 @@
-**Project detection**: the project's Verse path, name and root plugin now come from the .uefnproject owning the file being worked on, so a multi-root workspace holding two UEFN projects no longer answers every file from the first project.
+**Project detection**: the Verse path and root plugin now come from the .uefnproject owning the file being worked on, so a multi-root workspace with two UEFN projects no longer answers every file from the first.
--- a/changelog.d/450.fixed.md
+++ b/changelog.d/450.fixed.md
@@ -1,0 +1,1 @@
+**Project detection**: the project's Verse path, name and root plugin now come from the .uefnproject owning the file being worked on, so a multi-root workspace holding two UEFN projects no longer answers every file from the first project.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -352,7 +352,7 @@ export class ImportPathConverter {
         const workspaceFolder = vscode.workspace.getWorkspaceFolder(currentFileUri);
         if (!workspaceFolder) return;
 
-        const placement = ImportPathConverter.placeUnderContentRoot(workspaceFolder.uri.fsPath, currentFileUri.fsPath, await this.projectPathHandler.getRootPluginName());
+        const placement = ImportPathConverter.placeUnderContentRoot(workspaceFolder.uri.fsPath, currentFileUri.fsPath, await this.projectPathHandler.getRootPluginName(currentFileUri));
         if (!placement) return;
 
         // Every path below is reasoned about with a leading Content/, whatever
@@ -750,7 +750,7 @@ export class ImportPathConverter {
         const workspaceFolder = vscode.workspace.getWorkspaceFolder(documentUri);
         if (!workspaceFolder) return null;
 
-        const placement = ImportPathConverter.placeUnderContentRoot(workspaceFolder.uri.fsPath, documentUri.fsPath, await this.projectPathHandler.getRootPluginName());
+        const placement = ImportPathConverter.placeUnderContentRoot(workspaceFolder.uri.fsPath, documentUri.fsPath, await this.projectPathHandler.getRootPluginName(documentUri));
         if (!placement) return null;
 
         return placement.fileDirRelative ? `${projectVersePath}/${placement.fileDirRelative}` : projectVersePath;
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -868,7 +868,7 @@ export class ImportPathConverter {
             return null;
         }
 
-        const projectVersePath = await this.projectPathHandler.getProjectVersePath();
+        const projectVersePath = await this.projectPathHandler.getProjectVersePath(documentUri);
         if (!projectVersePath) {
             return null;
         }
@@ -1000,7 +1000,7 @@ export class ImportPathConverter {
 
         const { fullPath: modulePath, moduleName } = moduleInfo;
 
-        const projectVersePath = await this.projectPathHandler.getProjectVersePath();
+        const projectVersePath = await this.projectPathHandler.getProjectVersePath(documentUri);
         if (!projectVersePath) {
             vscode.window.showWarningMessage("Could not find .uefnproject file in workspace. Please ensure you have a valid UEFN project.");
             return null;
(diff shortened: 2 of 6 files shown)


vukefn:
Filed the pre-existing defect this change raises the stakes on: #461 - ImportPathConverter's own ProjectPathHandler never gets a watcher, so its cache is never cleared.

Issue #393, opened 2026-08-14
Use the qualified form for attributed re-declaration, without waiting on the compiler flag question


vukefn:
## What

The visibility chain writes a specifier-carrying re-declaration of an already-declared module — a second explicit definition — without establishing that the compiler accepts one. The compiler-sanctioned qualified form `(Parent:)Child := module:` declares into a scope without re-declaring it, and is safe whichever way the open flag question resolves.

## Where

- `visibilityResolution.ts:158-161` — the chain repeat that carries the keyword.
- `definitionsContent.ts:28-53` — the emission (`Name<spec> := module:` for nesting, `Name<spec> := module {}` for the leaf).
- The change lands in `buildDeclarationBlock`'s input.

## Effect

Reached whenever a deeper new module is written under a declared ancestor: publicizing `Systems/Combat/Weapons` where `Combat` is declared `<internal>` in a user file and `Weapons` is new. The retained chain then contains `Combat<public> := module:`, an attributed second explicit definition.

Error 3532 at the definitions file, in a `PublicUser` package without `treatModulesAsImplicit` — the project stops compiling on the exact edit that claimed to fix visibility. If UEFN sets the flag, the emission is legal: attribute levels match by construction, so no 3623.

The current design bets the flagship edit on a `.vpackage` setting nobody can read from source. The qualified form makes the bet unnecessary, so the defect is closable now rather than after the open question is answered.

## Evidence

- `SemanticAnalyzer.cpp:4438-4447` — the duplicate-definition gate, `PublicUser`-only. `Glitch.h:138` — 3532 `ErrSemantic_AmbiguousDefinition`.
- `SourceProject.h:115` — the flag defaults `= false`. `SourceFileProject.cpp:179` — it is read only from `.vpackage` JSON.
- The qualified form is compile-pinned: `PartialModuleDefinitions.versetest:168-184`, and `176-179` for mixed qualified-and-bare parts.
- The open question is not answerable from source: the only in-source knowledge of the flag being set is the digest generator's comment "This flag is always set for the asset manifest" (`DigestGenerator.cpp:161-167`), which concerns Epic's manifest package, not user packages; UEFN's user-package generator is outside the source drop; and the versetest harness sits outside the `PublicUser`-only check, confirmed by the commented-out 3532 block at `PartialModuleDefinitions.versetest:54-60` ("This is now detected via package VerseScope instead"). A live two-file compile remains the right experiment — but this writer need not wait for it.

## Smallest fix

[...]

PR #463, opened 2026-08-15
fix(visibility): report a re-declared ancestor instead of writing it

### LINKED ISSUE #449: Visibility writer repeats an explicitly-declared ancestor in the chain without surfacing it as a conflict
The module visibility quick fix writes a nested declaration chain into the Content-root definitions file. Where every module in the chain is new, this is correct and nothing here applies. The defect is the other case: an ancestor that already carries an explicit declaration in a user file. The writer repeats it as a second explicit definition, and the user is told nothing.

Rendered from the compiled `buildDeclarationBlock` for the chain `visibilityResolution` returns when `Gadgets<internal> := module:` is already declared in another file:

```
Gadgets<internal> := module:
    Tools<public> := module {}
```

That output is pinned by a checked-in test, `src/visibility/__tests__/visibilityResolution.test.ts:97-105` ("repeats an existing parent specifier in the written chain, so the parts cannot disagree"), so it is the intended behaviour today rather than an accident.

Root cause, 2 parts:

1. `src/visibility/visibilityResolution.ts:160-161` pushes a chain entry for a module that already has a declaration, carrying its keyword. The attribute is not what makes it dangerous: `:136` pushes a bare repeat on the blocked path, and `:161` emits bare whenever the found declaration carries no specifier. All three shapes are second definitions.
2. `src/visibility/definitionsContent.ts:28-35` renders every chain entry as a full `Name<spec> := module:` definition, so a repeat becomes a second *explicit* definition of a module the project already defines explicitly.

Why the repeat is a risk: `SemanticAnalyzer.cpp:4437-4447` rejects a second explicit module definition in a `PublicUser` package with 3532 (`Glitch.h:138`) unless `_bTreatModulesAsImplicit` is set on either package. Whether UEFN sets that flag on user packages is not answerable from the source drop — `SourceProject.h:115` defaults it false and `SourceFileProject.cpp:179` reads it only from `.vpackage` JSON.

Not at risk: a chain nesting only through folder-only modules. The gate needs the *existing* definition to be explicit (`:4437`), and `visibilityResolution.ts:124` routes a module with no declaration down the new-module branch instead.

Severity: the edit that claims to fix visibility can stop the project compiling, on any project where an ancestor is explicitly declared, whenever a deeper module is publicized under it. Everything else works — all-new chains, and the in-place specifier edits.

Fix direction:

- Surface the repeat through the conflict channel that already exists. `visibilityResolution.ts:130-139` collects `blocked` entries and the function returns them as `conflicts` with a `widen`/`nest` reason; this case wants the same treatment with its own reason, so the quick fix can say "this ancestor is explicitly declared and the chain would have to repeat it" instead of silently writing a file that may not compile.

One decision belongs to the implementing session, and it should be made deliberately rather than fallen into: whether the new conflict **refuses** the write, as `scoped` does, or is **reported while the write still proceeds**. Refusing everywhere would regress the quick fix on projects where the flag is set and the repeat is legal; proceeding keeps today's risk but removes the silence. The conservative reading is that making the situation visible is the whole of what is flag-independent, and the refuse-or-proceed choice can follow the flag answer later.

Out of scope, deliberately:

- Whether the repeat is legal at all — the `treatModulesAsImplicit` question. It needs a live two-file compile in a real UEFN project, and nothing in this ticket depends on the answer.
- Any redesign of where the writer puts a new module's declaration. The qualified form `(Parent:)Child := module:` is not a route to it: `SemanticAnalyzer.cpp:17873-17881` requires the qualifier to name the definition's own enclosing module, and `ModuleVisibilityWriter.ts:330-337` always places the definitions file at the Content root.

Background, including why the qualified form was ruled out: #393.

vukefn:
Closes #449

## What was wrong

The declaration chain restated any ancestor that already carried an explicit declaration, and a restatement is a second explicit definition. `SemanticAnalyzer.cpp:4437-4447` rejects one in a `PublicUser` package with `ErrSemantic_AmbiguousDefinition` (3532, `Glitch.h:138`) unless `_bTreatModulesAsImplicit` is set on either package — so the edit that claimed to fix a visibility error could stop the project compiling, silently.

The most reachable form is not the one the ticket describes. The definitions file's own declarations feed back into the scan (`ModuleVisibilityWriter.ts:441-443`), so the **second** run of the quick fix in one subtree found the first run's output and appended a block defining the same modules again. That was pinned as intended behaviour by a checked-in test, whose expected value was:

```
Gadgets := module:
    Deep<public> := module {}

Gadgets := module:
    Deep<public> := module:
        Tools<public> := module {}
```

One file, `Gadgets` defined twice and `Deep` defined twice.

## The rule, and how it was settled

The gate does **not** depend on whether the new part carries attributes: `bTreatAsImplicit = !bHasAttributes && bPackageTreatsModulesAsImplicit` (`:4428`) feeds only `CreatePart` at `:4458`, never the gate. So bare repeats carry the same exposure as attributed ones, which is why all three shapes the ticket names are covered rather than just the keyword-carrying one.

The narrowing in the other direction holds and is respected: the gate needs the *existing* definition to be explicit (`:4437`), and `visibilityResolution.ts` already routes an undeclared module down the new-module branch. A chain nesting only through folder-only modules is not at risk and is **not** reported — the ordinary first-use quick fix is unchanged.

Checked against the shipped compiler, not only `ue5-main`: the diagnostic templates are byte-identical in `epicgames.verse-0.0.56430492/bin/Win64/verse-lsp.exe`.

## The decision the ticket deferred

#449 left it to the implementing session whether the new conflict refuses the write or is reported while the write proceeds. **Refuse**, because `conflicts` is a hard gate by construction — `ModuleVisibilityWriter.ts` returns a `Refusal` before any `WorkspaceEdit` is built — so routing through the channel the ticket names *is* refusing, and building a report-and-proceed channel would be the design change both #393 and #449 put out of scope. It is also the cheap direction to reverse if the `treatModulesAsImplicit` question later resolves favourably: lifting the refusal deletes one filter.

Accepted cost: a project with an explicitly-declared ancestor now gets a refusal where it previously got a write that may have been legal. The in-place specifier path — the common "just make this module public" case — is untouched.

## Changes

- `visibilityResolution.ts` — `VisibilityConflict.reason` gains `repeat`; `keyword` becomes optional, since a repeated ancestor may be declared bare. Declared modules are collected alongside the existing `blocked` and filtered by the same `chainIndex <= deepestWritten` trim, so a repeat that gets trimmed out of the retained chain is not reported.
- `ModuleVisibilityWriter.ts` — `refusalForConflicts` gains the `re-declaring` clause. The keyword is bracketed and lists take a serial comma, because one conflict per declared ancestor makes multi-entry lists ordinary and the old join fused a keyword clause with the path after it.

`definitionsContent.ts` is untouched, and so is where the writer puts a new module's declaration.

## Verification

```
$ npm run compile
> tsc -p ./
```

```
$ npm test
Test Suites: 52 passed, 52 total
Tests:       1589 passed, 1589 total
```

```
$ npm run format:check
Checking formatting...
```

## Review

Two rounds at `opus`, both `PASS_WITH_SUGGESTIONS`, no `Critical` in either.

Round 1 (3 Important) and round 2 (1 Important) were all acted on: the refusal message was unparseable at three or more entries; the reported keyword depended on arbitrary scan order for a module declared in several files; and two doc comments asserted rules the change had made dead.

Both reviewers independently confirmed the load-bearing claim that a specifier copied off an existing declaration never reaches a file — the trim and the `repeat` test are literally the same predicate — and that the ordinary folder-only quick fix still writes.

## Known consequence, left alone deliberately

The definitions-file specifier fold (`ModuleVisibilityWriter.ts`, the `inDefinitions` branch) is now unreachable: any specifier edit against that file implies its module's ancestors are declared there too, which makes the module itself a `repeat`. Both reviewers proved this independently and probed six shapes, all refusing.

I left the branch in place rather than deleting it — it is the guard that stops overlapping edits, and it goes live again the moment the refusal is relaxed. Worth a decision from you, not from me.

## Follow-up worth filing

Teaching the writer to nest a new declaration **inside** the block it already wrote, instead of appending a second one, would turn the most common refusal back into a working quick fix. That is the "where the writer puts things" redesign #393 and #449 both exclude, so it is not in this PR. I have not filed it — say the word and I will.

### DIFF
--- a/changelog.d/449.fixed.md
+++ b/changelog.d/449.fixed.md
@@ -1,0 +1,1 @@
+**Module visibility**: the make-module-public quick fix now refuses, naming the module, where the declaration chain would re-declare an ancestor your project already declares, instead of writing a duplicate definition that can stop the project compiling.
--- a/src/visibility/ModuleVisibilityWriter.ts
+++ b/src/visibility/ModuleVisibilityWriter.ts
@@ -543,13 +543,20 @@ export class ModuleVisibilityWriter {
  * remedy before it is read.
  */
 function refusalForConflicts(moduleName: string, conflicts: readonly VisibilityConflict[]): string {
+    // Bracketed rather than trailing a comma, and the list punctuated: one
+    // conflict per declared ancestor is ordinary, and a reader given
+    // `A and B, declared internal and C` cannot tell where a path ends.
+    const joined = (items: readonly string[]) => (items.length <= 2 ? items.join(" and ") : `${items.slice(0, -1).join(", ")}, and ${items[items.length - 1]}`);
+
     const listed = (reason: VisibilityConflict["reason"]) =>
-        conflicts
-            .filter((conflict) => conflict.reason === reason)
-            // A repeat of a declaration carrying no specifier has no keyword to
-            // name, and `declared undefined` would read as one it does carry.
-            .map((conflict) => (conflict.keyword ? `${conflict.path}, declared ${conflict.keyword}` : conflict.path))
-            .join(" and ");
+        joined(
+            conflicts
+                .filter((conflict) => conflict.reason === reason)
+                // A repeat of a declaration carrying no specifier has no
+                // keyword to name, and `declared undefined` would read as one
+                // it does carry.
+                .map((conflict) => (conflict.keyword ? `${conflict.path} (declared ${conflict.keyword})` : conflict.path)),
+        );
 
     const widen = listed("widen");
     const nest = listed("nest");
--- a/src/visibility/ModuleVisibilityWriter.ts
+++ b/src/visibility/ModuleVisibilityWriter.ts
@@ -67,12 +67,14 @@ export class ModuleVisibilityWriter {
      * Puts the `<public>` declarations that make the requested module
      * reachable through the refactor preview, and reports what came back.
      *
-     * Refuses rather than half-applying: a module declared anything but
+     * Refuses rather than half-applying, and everything else the same request
+     * would have changed goes with it. A module declared anything but
      * `<public>` or `<internal>` is left alone, whether it is the one to
-     * publicize or one a new declaration would sit inside, and so is everything
-     * else the same request would have changed. That governs what is offered,
-     * not what lands - the preview lets the user apply a subset, so what is
-     * reported afterwards is intent rather than outcome.
+     * publicize or one a new declaration would sit inside; so is one the chain
+     * would have to declare a second time, since the block is appended whole
+     * and cannot join a declaration the project already carries. That governs
+     * what is offered, not what lands - the preview lets the user apply a
+     * subset, so what is reported afterwards is intent rather than outcome.
      *
      * @param sourceUri the document whose diagnostic asked for this, which is
      * what picks the workspace folder in a multi-root workspace. The first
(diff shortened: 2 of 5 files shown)


vukefn:
Follow-up filed as #464: nesting a new declaration into the block the writer already wrote, which turns the most common refusal this PR introduces back into a working quick fix.

PR #465, opened 2026-08-15
fix(imports): resolve the Content root at any depth for project-wide scans

### LINKED ISSUE #448: Content root read at a fixed depth in the scanning and visibility paths
## What

Four sites still locate the project's `Content` root by reading it at a fixed depth from the workspace folder, so a project that nests its Verse under `Plugins/<Name>/Content` is invisible to all of them. #429 fixed the two sites on the document-relative path; these are the rest, and they now disagree with the helper that replaced them.

## Where

- `ImportPathConverter.ts` — `workspaceScanTargets`, which globs `Content/**/*.verse` from the workspace folder. Module-search phases 2 and 3 both read it, so neither can see a nested project.
- `ProjectPathCache.ts` — `workspaceIsContent()`, on the same rule, so every cached declaration candidate for a nested project is dropped.
- `moduleLocationLookup.ts` — `toContentRelativeDir`, which maps a found file back to a Content-relative directory and returns nothing for a path that does not start with `Content/`.
- `ModuleVisibilityWriter.ts` — `findContentRoot`, the same boundary for a different feature: "Make module public" answers `no 'Content' folder was found in the workspace` in exactly the layout import conversion now handles.

## Effect

Probed on the #429 branch, with the workspace opened at the project root and Verse under `Plugins/<Name>/Content`: "Use full path" on a module that only the project-wide scan could reach still fails with `Could not find module`. The `findFiles` pattern issued is `Content/**/*.verse` against the project root, which matches nothing. Phase 1 walks outward from the document and now succeeds, which is what made #429's case work, so the failure is confined to targets phase 1 cannot reach.

Fail-closed in every case, so nothing writes a wrong import — this is a missing capability rather than a correctness defect. The visibility writer's refusal is loud and actionable; the search sites are silent.

## Why this is not #396

#396 covers the byte-exact `Content` comparison (H-08, the *case* rule: an on-disk `content` against a `Content` in code) and scanning `workspaceFolders[0]` rather than the folder owning the document (H-14). Neither is the *depth* rule, and fixing either would leave every site above unchanged. Filed separately so the depth rule has one owner across all four.

## Smallest fix

`ImportPathConverter.placeUnderContentRoot` already resolves a Content root at any depth from a document, gated on the project file's root plugin. The scanning sites have no document to walk from, so they need the root derived once per operation instead — from the `.uefnproject` location plus `plugins[bIsRoot].name`, which `ProjectPathHandler.getRootPluginName()` now exposes — and then threaded through the glob, the cache validity check and the map-back.

Worth doing as one change rather than four: the reason #429 existed is that two sites read this boundary two ways, and four more reading it a third way is the same shape.

---
**Source:** round-2 review of PR #447 (#429), Important finding, confirmed by probe.

vukefn:
Closes #448

## What

#429 taught the two document-relative sites to find a Content root nested under `Plugins/<Name>`. The four sites with no document to walk from still read the boundary at a fixed depth off the workspace folder, so a workspace opened at the project root was invisible to the declaration scan, the folder scan and the cache lookup, and "Make module public" refused it with `no 'Content' folder was found in the workspace`.

The root is now resolved once per operation, as an absolute URI, and threaded through the glob, the cache lookup and the map-back.

## How

**`src/services/contentRoot.ts`** (new) states the boundary once, in the same terms `ImportPathConverter.placeUnderContentRoot` uses:

- `contentRootCandidates` — pure, shallowest first: `""` when the workspace folder is itself named `Content`, otherwise `Content`, then `Plugins/<rootPluginName>/Content` when the project file declares a root plugin.
- `findContentRoot` — the first candidate that stats as a directory.

**`moduleLocationLookup`** — `toContentRelativeDir(sourceFileAbsolute, contentRootAbsolute)` replaces the `workspaceIsContent` boolean, and `resolveModuleLocations` takes `{ workspaceFolderPath, contentRootPath }`. Absolute paths let `path.relative` subtract the root instead of a caller stripping a prefix it spelled itself, which matters because the nested candidate is composed from the plugin name the project file declares and that name is matched against a directory rather than resolved, so it need not match its case. Relating two absolute paths folds exactly as far as the `stat` that found the root did.

**`ImportPathConverter`** — `workspaceScanTargets` is gone; both scan phases resolve the root once and glob `RelativePattern(contentRoot, "**/*.verse")`.

**`ProjectPathCache.lookupModuleLocations`** — async, and resolves the root before placing a cached `sourceFile`.

**`ModuleVisibilityWriter`** — `findContentRoot` and `contentRelativeDirectory` are gone. `buildEdit` passes the located root into `findDeclarations`, so the folder written to and the folder read from can no longer come apart: the rule was previously derived three times inside one operation.

## Why this shape

Settled against real UEFN output rather than off this extension's own scanner:

- A UEFN project root holds only `<Project>.uefnproject` and `Plugins/`. There is no project-level `Content` — checked across every project under a local UEFN project directory at depth 2, zero hits. So shallowest-first candidate ordering never competes with the root plugin's Content on a real layout, and this introduces no divergence from #429's rule.
- UEFN's generated `.code-workspace` opens folder 0 *at* `Plugins/<Name>/Content`. That is the depth-0 case the fixed-depth rule was written for, and why the defect only surfaces for someone who opens the project root instead.

## Out of scope

Which workspace folder is chosen (#396 H-14, #450) and the byte-exact `Content` comparison (#396 H-08) are untouched, as the ticket asks.

## From the review gate

Two findings were fixed rather than only reported, both pinned by tests that fail without them:

- **`FileStat.type` is a bitmask.** The disk provider ORs `SymbolicLink` onto the target's type, so `stat.type === FileType.Directory` rejects a Content root reached through a symlink or a junction. These phases globbed with no stat at all before and `search.followSymlinks` defaults to true, so a junctioned root worked and would have stopped working. The pre-existing `ImportPathConverter.folderExists` carries the same equality test; left alone as out of scope.
- **The root plugin name is joined into a path.** A project file carrying a separator or a relative segment in it could root the scan outside the workspace folder, where the visibility writer would then create its definitions file. Refused the way `definitionsUri` already refuses a file name carrying a path.

One finding is reported and not fixed, deliberately:

- **The nested candidate is composed against the workspace folder, not the `.uefnproject` directory.** The ticket's "Smallest fix" names the project-file location as the anchor; only `getRootPluginName()` was used. The gap is a workspace opened at `<project>/Plugins`, where `placeUnderContentRoot` still places a document but the scans find no root — a layout UEFN does not generate. Closing it means anchoring on `path.dirname(projectFilePath)`, which needs `ProjectPathHandler.findAndParseProjectFile` to retain the path it currently discards — the same method #450 is restructuring. Worth a follow-up once #450 lands rather than a conflicting edit now.

Also noted, not acted on: `CONTENT_FOLDER` is declared in three files and `PLUGINS_FOLDER` in two; `placeUnderContentRoot` still encodes the same two rules in its own shape (it walks a known file path where this probes a directory); and the root is resolved per call rather than memoized per operation.

## Tests

Regression tests at the entry points the issue names. Six pin the depth rule and were proved to detect it — disabling the nested candidate fails exactly those six and nothing else:

- `ImportPathConverter.findModuleLocations` through phase 2 (declaration scan) and phase 3 (folder scan) under a nested root, the glob's base, and the fail-closed case where no root plugin is declared.
- `ProjectPathCache.lookupModuleLocations` placing a nested `sourceFile`, the flat case still working, and the no-root-plugin case.
- `ModuleVisibilityWriter` writing at the nested root, rewriting a declaration the scan finds there, and still refusing when no root plugin is declared.

Two more pin the review fixes, each likewise proved to fail without them: a Content root reported as `Directory|SymbolicLink`, and `contentRootCandidates` refusing a plugin name that carries a separator or a relative segment. The candidate list is tested directly rather than through a scan because the mock's `Uri.joinPath` does not resolve `..`, which made the end-to-end version of that test pass for the wrong reason.

## Verification

```
> tsc -p ./
```

```
> jest --verbose

Test Suites: 54 passed, 54 total
Tests:       1607 passed, 1607 total
Snapshots:   0 total
Time:        8.15 s
Ran all test suites.
```

```
> prettier --check "**/*.ts"

Checking formatting...
```

### DIFF
--- a/changelog.d/448.fixed.md
+++ b/changelog.d/448.fixed.md
@@ -1,1 +1,1 @@
-**Module search and visibility**: project-wide module lookups and "Make module public" now find the project's Content folder wherever UEFN nests it, so a workspace opened at the project root works the same as one opened at Content.
+**Module search and visibility**: project-wide module lookups and "Make module public" now find a Content folder nested under `Plugins/<Name>`, so a workspace opened at the project root works the same as one opened at Content.
--- a/changelog.d/448.fixed.md
+++ b/changelog.d/448.fixed.md
@@ -1,0 +1,1 @@
+**Module search and visibility**: project-wide module lookups and "Make module public" now find the project's Content folder wherever UEFN nests it, so a workspace opened at the project root works the same as one opened at Content.
--- a/src/imports/ImportPathConverter.ts
+++ b/src/imports/ImportPathConverter.ts
@@ -3,6 +3,7 @@ import * as path from "path";
 import { logger } from "../utils";
 import { ProjectPathHandler } from "../project";
 import { ProjectPathCache } from "../services";
+import { findContentRoot } from "../services/contentRoot";
 import { buildProjectIndexes, resolveFolderModuleLocations, resolveModuleLocations, toContentRelativeDir } from "../services/moduleLocationLookup";
 import { ProjectPathNode } from "../types";
 import { findExplicitModuleDeclarations } from "../visibility/moduleDeclarations";
@@ -418,20 +419,17 @@ export class ImportPathConverter {
     }
 
     /**
-     * Where a workspace-wide `.verse` scan has to look, and whether the
-     * workspace folder is itself the Content folder - which decides both the
-     * glob and how a found path maps back to a Content-relative one.
+     * The Content root a workspace-wide `.verse` scan has to look under, or
+     * null when the project has none there.
      *
-     * Shared by the two scanning phases so they cannot disagree about the
-     * layout: a phase reading one rule and a phase reading the other would
-     * resolve the same project differently.
+     * Neither scanning phase has a document to walk out from, so the root is
+     * resolved from the workspace folder and the project's root plugin instead
+     * of read at a fixed depth. Shared by both phases so they cannot disagree
+     * about the layout: a phase reading one rule and a phase reading the other
+     * would resolve the same project differently.
      */
-    private static workspaceScanTargets(workspaceFolder: { uri: vscode.Uri }): { searchPattern: string; workspaceIsContent: boolean } {
-        const workspaceIsContent = path.basename(workspaceFolder.uri.fsPath) === CONTENT_FOLDER;
-        return {
-            searchPattern: workspaceIsContent ? "**/*.verse" : `${CONTENT_FOLDER}/**/*.verse`,
-            workspaceIsContent,
-        };
+    private async scanContentRoot(workspaceFolder: { uri: vscode.Uri }): Promise<vscode.Uri | null> {
+        return findContentRoot(workspaceFolder, await this.projectPathHandler.getRootPluginName());
     }
 
     /**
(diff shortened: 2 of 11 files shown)


vukefn:
Follow-up for the anchor gap noted above: #466, blocked on #450.

PR #471, opened 2026-08-16
fix(imports): take the consumer override only from a diagnostic inside the using clause

### LINKED ISSUE #412: A diagnostic on a statement sharing a pinned using line can license placing an import above its provider
## What

`couldResolveAgainst` takes its evidence as a line, not a range. A pinned import that shares its line with another statement via `;` therefore takes the consumer override from a diagnostic about that *other* statement — and the override places the new import above it, past the provider it may genuinely need.

## Where

- `src/imports/ImportDocumentEditor.ts` — `couldResolveAgainst`, the span test `line >= imp.startLine && line <= imp.endLine`. Its own doc already states the trade-off: "A line, not a range, so an import pinned by a statement sharing its line still takes the override from a diagnostic about that other statement. Narrowing it needs a column, which `ScannedImport` does not carry."

## Reproduction

```verse
using { Economy.Shop }; Thing := Features.MakeIt()
```

Add `Features` with a diagnostic reported on line 0. The diagnostic is about the code after the `;`, not about the `using`, but the span test cannot tell them apart. Result:

```verse
using { Features }

using { Economy.Shop }; Thing := Features.MakeIt()
```

If `Economy.Shop` is what brings `Features` into scope, the rewritten file stops compiling.

## Scope

Both entry points, and pre-existing on both. `addImportsToDocument` produces the identical insert on the identical input today; #385 (PR #411) extended the same accepted trade-off to `organizeImports`, which is what that ticket asked for. This is not a regression from either.

## Why it is narrow

The override needs a rank-2 pinned import, a rank-0 or rank-1 new path, and a diagnostic landing inside the pinned statement's span. A file that reaches it is one where a `;` puts an unrelated statement on a `using` line and the compiler reports on that statement.

## Direction

`ScannedImport` carries no column, so the span test cannot currently distinguish a diagnostic inside the `using` clause from one in the statement beside it. Recording the `using` statement's column range at scan time would let `couldResolveAgainst` ask whether the diagnostic falls inside the clause itself.

Worth weighing against the cost: the failure direction here is a file that stops compiling, which `hoistFloor` elsewhere treats as outranking a fixed diagnostic. That asymmetry is the argument for narrowing it rather than leaving the trade-off symmetric.

---
**Severity:** LOW — narrow trigger, pre-existing, documented at the call site.
**Source:** review gate on PR #411 (issue #385), probe "can the evidence place an import above a pinned import that genuinely provides for it".

vukefn:
Closes #412

## What

`couldResolveAgainst` read its evidence by line, so a diagnostic about a statement sharing a `;`-joined line with a pinned `using` licensed the consumer override, and the new import was written above a provider the pinned import may genuinely need — a compiling file stopped compiling.

The scanner now records the raw column span of a statement that shares its single line with other code (`ScannedImport.columns`), measured on the length-preserving `masked` string and reconciled against the code-derived path sequence — any disagreement leaves the span absent, falling back to today's line test rather than measuring against the wrong statement. Diagnostic evidence carries the range's start position instead of the line alone (`DiagnosticPositionsByPath` / `DiagnosticPositionsByStatement`), recorded by all three producers. `couldResolveAgainst` requires the position to start inside the clause where a span is recorded.

Multi-line pinned spans (a trailing `using:` pair over its indented path) keep line-granular evidence — the same class of gap one shape along, documented at the predicate and left for a follow-up rather than half-closed here.

## Domain rules relied on

- `using { X }; Y := 1` on one line is legal Verse; `;` separates definitions as a newline does (maintainer live-compiler check, 2026-08-10).
- The missing-`using` / unknown-identifier glitch anchors to the identifier's own VstNode (`SemanticAnalyzer.cpp:12544`, `:12864`), so a diagnostic's start character lands inside the text it is about.
- Degradation is one-sided: a whole-line-range diagnostic reproduces today's behavior or refuses the override, and a refused override leaves a diagnostic unfixed where a wrong override breaks a compiling file — the asymmetry `hoistFloor` already encodes.

## Verification

`npm run compile` — exit 0.

`npm test`:

```
Test Suites: 57 passed, 57 total
Tests:       1649 passed, 1649 total
```

`npm run format:check`:

```
Checking formatting...
```

Regression tests are pinned at both entry points (`addImportsToDocument` and `organizeImports`/`buildOrganizedContent`); with the predicate reverted to the line test, the three defect-direction tests fail and the override-retained tests pass.

### DIFF
--- a/changelog.d/412.fixed.md
+++ b/changelog.d/412.fixed.md
@@ -1,0 +1,1 @@
+**Import placement**: a new import is no longer forced above a `using` sharing its line with another statement when the compiler's error was about that other statement, which could stop the file compiling.
--- a/src/commands/CommandsHandler.ts
+++ b/src/commands/CommandsHandler.ts
@@ -4,6 +4,7 @@ import { ImportHandler, ImportPathConverter, ImportCodeLensProvider } from "../i
 import { StatusBarHandler } from "../ui";
 import { ProjectPathCache } from "../services";
 import { ModuleVisibilityRequest, ModuleVisibilityWriter } from "../visibility";
+import { DiagnosticPosition } from "../types";
 
 /** The collaborators the commands act on, wired once during activation. */
 export interface CommandsDependencies {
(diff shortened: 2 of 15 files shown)


vukefn:
The review gate confirmed one residual shape of the same class, deliberately scoped out here and documented at the predicate: a pinned multi-line `using:` pair still takes line-granular evidence. Filed as #472.

PR #474, opened 2026-08-16
fix(imports): take the pair override only from evidence inside its span

### LINKED ISSUE #472: A diagnostic beside a pinned multi-line using: pair still licenses placing an import above its provider
## What

#412 narrowed the pinned-import consumer override to the `using` clause's own columns, but only for statements whose span is a single line. A pinned multi-line pair — `X := 1; using:` over its indented path — carries no `columns`, so `couldResolveAgainst` still reads its evidence line-wide: a diagnostic about `X := 1` on the opener line takes the pair's override and places the new import above a provider the pair's path may genuinely need. The same failure direction as #412 (a compiling file stops compiling), one shape further along.

## Where

- `src/imports/ImportDocumentEditor.ts` — `couldResolveAgainst`, the no-`columns` fallback branch; its doc names this as a standing gap.
- `src/imports/ImportScanner.ts` — `ScannedImport.columns` is present only with `startLine === endLine`, so the pinned-pair push sites cannot record what the predicate would need.

## Reproduction

Confirmed by the #412 review gate's probe:

```verse
X := 1; using:
    Economy.Shop
```

Evidence for a new bare path at position `(0, 0)` — inside `X := 1`, outside the `using:` clause — still hoists the new import above the pinned pair.

## Why it is narrow

The trigger needs a trailing `using:` pair with a rank-2 (dotted) path, a rank-0/1 new path, and a diagnostic landing in the statements before the opener. Rarer than the single-line shape #412 fixed.

## Direction

The span the predicate needs is two positions, not one line-local pair: from the opener's column on `startLine` to the end of the path on `endLine`. Either widen `columns` to carry line-qualified endpoints for multi-line spans, or record a second field for the pair shape. The opener column is measurable on the masked line the scanner already carries (`LineClassification.masked`); the path line's extent is the path text before its trailing comment.

---
**Severity:** LOW — narrow trigger, documented at the call site.
**Source:** review gate on PR #471 (issue #412), probe "multi-line pinned pair still takes line-granular evidence".

vukefn:
Closes #472

## What

`couldResolveAgainst` read a pinned multi-line pair's evidence line-wide, so a diagnostic about the statement beside the opener — `X := 1` on `X := 1; using:` — took the consumer override and the new import was written above a provider the pair's dotted path may genuinely need: a compiling file stopped compiling, the #412 failure one shape further along.

The scanner now records the span endpoints the predicate needs (`ScannedImport.spanColumns`: the `using` opener's column on `startLine`, the offset just past the path on `endLine`), measured on the length-preserving `masked` strings and reconciled against the code-derived reading that admitted the pair — either side disagreeing leaves the field unset and the predicate on today's whole-span test, the same one-sided degradation `columns` takes. `couldResolveAgainst` requires opener-line evidence to start at or after the opener and endLine evidence to start before the path's end; lines strictly inside the span count whole, since the pair tolerates a blank or comment line there and such a line holds no other statement for the evidence to be about.

The multi-line braced clause keeps line-granular evidence — its paths can split across lines, so it has no one line to measure an endpoint on. Named at the predicate as the remaining loose shape.

## Domain rules relied on

- A missing-`using` diagnostic anchors to the identifier's own VstNode, so its start position lands inside the statement it is about (`CreateGlitchForMissingUsing` call sites, SemanticAnalyzer.cpp:12544, :12864).
- The indented clause's body is the first line of code below the opener; neither a blank line nor a comment ends it, and the clause holds exactly one expression (`AnalyzeMacroCall_Using`, SemanticAnalyzer.cpp:6378-6381; the `Line` production's unconditional `Gen.BlankLine`, VerseGrammar.h:1683).
- `X := 1; using:` over an indented path is legal Verse — `;` separates definitions as a newline does (maintainer live-compiler check, 2026-08-10).

## Verification

`npm run compile` — exit 0.

`npm test`:

```
Test Suites: 58 passed, 58 total
Tests:       1696 passed, 1696 total
```

`npm run format:check`:

```
Checking formatting...
```

Written red-first: with the predicate and scanner unfixed, the three defect-direction tests fail (override still taken from evidence beside the opener at both entry points, and from evidence past the path) and the override-retained tests pass.

### DIFF
--- a/changelog.d/472.fixed.md
+++ b/changelog.d/472.fixed.md
@@ -1,0 +1,1 @@
+**Imports**: A diagnostic about a statement written beside a pinned `X := 1; using:` pair no longer places the new import above the provider the pair's dotted path may need.
--- a/src/imports/ImportDocumentEditor.ts
+++ b/src/imports/ImportDocumentEditor.ts
@@ -473,12 +473,20 @@ export class ImportDocumentEditor {
      * the line, outside a clause written after a `;`, which refuses the
      * override - the same safe direction.
      *
-     * Without `columns` the whole span still counts. That keeps a statement
+     * A multi-line pair carries `spanColumns` instead, and the evidence must
+     * start at or after its opener: from there to the end of the path on the
+     * last line, everything is the clause, and a diagnostic before the opener
+     * is about the statement written beside it - the same shared-line shape
+     * one line taller. Lines strictly inside the span count whole; the pair
+     * tolerates a blank or comment line there, and such a line holds no other
+     * statement for the evidence to be about.
+     *
+     * Without either field the whole span still counts. That keeps a statement
      * owning its span exactly as sharp as before, and leaves one shape loose:
-     * a pinned multi-line span - `X := 1; using:` over its indented path -
-     * still takes the override from a diagnostic about the statement beside
-     * its opener. A standing gap rather than one this predicate closes;
-     * narrowing it needs a span the scanner does not measure.
+     * a pinned multi-line braced clause still takes the override from a
+     * diagnostic about a statement beside its opener. Its paths can be split
+     * across lines, so it has no one line to measure an endpoint on, and the
+     * span goes unrecorded.
      *
      * Still kept to a pinned dotted consumer and a new import that can supply a
      * first segment for it, because the override is only worth the risk that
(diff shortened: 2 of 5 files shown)


vukefn:
The review gate's probes surfaced two braced-clause gaps outside this ticket's scope, now filed: #475 (a statement after the clause's closing brace still takes the consumer override) and #476 (a multi-line braced clause opened after a statement is never scanned). The predicate's doc names the first as the remaining loose shape.

PR #480, opened 2026-08-16
fix(project): watch the project file where it lives, through one owner

### LINKED ISSUE #376: Project identity: one owner emitting one event, replacing three watchers

vukefn:
Closes #376

## What was wrong

Project identity was watched by three independent `createFileSystemWatcher("**/*.uefnproject")` string-glob watchers (`ProjectPathHandler`, `ProjectPathCache`, `AssetsDigestParser`), with only a timing coincidence ordering their reactions. A string glob reports no events from outside the opened workspace folders, and in the UEFN-generated multi-root workspace the `.uefnproject` sits outside every folder — so a mid-session project change (a rename, UEFN rewriting bindings) was never observed: stale identity served for the rest of the session, a cache that never rebuilt, stale asset class names. Discovery's parent walk also leaned on `findFiles` with an out-of-workspace `RelativePattern` base, which the API documents only for workspace-folder search.

## What changed

- **`src/project/ProjectPathHandler.ts`** — now the one identity owner. It exposes `onDidChangeProject`, fired after its cache clear (clear-before-fire is the ordering contract the three components previously prayed for). The string-glob watcher stays for the single-root case; a project directory discovered **outside** the workspace folders gets its own watcher anchored with `RelativePattern(Uri.file(dir), "*.uefnproject")` — the documented out-of-workspace watching form, and the same shape the digest watcher already used. Anchoring runs from the discovery success path and, for discoveries that finished before watcher setup, at `setupFileWatcher()` itself. The parent walk probes directories with `workspace.fs.readDirectory` instead of `findFiles`.
- **`src/services/ProjectPathCache.ts`** — drops its own `.uefnproject` watcher; the rebuild subscribes to `onDidChangeProject`. Deliberate behavior change: the shared event also fires on **delete** (the old watcher subscribed change/create only); the scan then finds no project and commits nothing, leaving the cache cleared rather than serving the deleted project's paths. `getStats()` now reports `projectName`.
- **`src/services/AssetsDigestParser.ts`** — drops its own `.uefnproject` watcher; the debounced refresh subscribes to the same event. The "nothing contracts that order" admission on `REFRESH_DEBOUNCE_MS` is gone because the order is now contracted.
- **`src/commands/CommandsHandler.ts`** — `rebuildPathCache` returns the post-rebuild stats, so `executeCommand` relays them to the extension-host test.
- **Tests** — new `ProjectPathHandler.watcher.test.ts` (string-glob event, anchoring, no-anchor-inside-workspace, the activation-order race, disposal); new `projectIdentity.test.ts` wiring the real trio exactly as `extension.ts` does and asserting a rename outside the workspace folders reaches identity, cache and digest; the extension-host T7 now asserts the rebuilt cache reports `loaded: true` with the fixture project's name, pinning the whole discovery chain in a real VS Code host (the fixture `.uefnproject` sits above every opened folder). Existing watcher-shape tests updated to the event-driven wiring. Mutation check: with anchoring disabled, 4 of the 6 new tests fail; the 2 survivors are the string-glob cases that rightly don't depend on it.

Known gap, per the ticket's own "after discovery succeeds" prescription: a `.uefnproject` **created** outside the workspace folders in a session whose discovery never succeeded is picked up lazily by the next identity read, not by a watcher — the subscriber comment now says so instead of promising otherwise.

Out of scope, follow-up candidate: `ImportPathConverter` (and `ImportCodeLensProvider`'s own converter) construct private `ProjectPathHandler` instances that no watcher or event ever invalidates.

## Verification

`npm run compile` — clean.

`npm test`:

```
Test Suites: 65 passed, 65 total
Tests:       1718 passed, 1718 total
Snapshots:   0 total
Time:        11.554 s
```

`npm run format:check` — "All matched files use Prettier code style!"

`npm run test:integration` (extension-host suite):

```
  commands and snooze (playbook T7/T8)
    - T7/T8: every contributed command is registered
    - T7: rebuild path cache discovers the fixture project outside the workspace folders (124ms)
 30 passing (33s)
```

## Review

One round at the configured review tier: **PASS_WITH_SUGGESTIONS**, 0 Critical, 1 Important, 5 Suggestions, 28 behavioural probes. The Important (a comment still promising create-after-startup recovery the UEFN shape cannot deliver) and three suggestions (two stale watcher comments, the emitter one-call contract, the case-sensitivity mismatch between the two discovery paths) are fixed in the second commit. Kept deliberately: the sorted multi-file pick (replaces platform-dependent `readDirectory` order) and the plain `{dispose}` handle (it flips the anchoring flag, which `Disposable.from` cannot express alone).


vukefn:
Follow-up filed for the identity holders this PR deliberately leaves untouched: #481 (the private ProjectPathHandler instances inside ImportPathConverter and ImportCodeLensProvider's own converter).

